PR #2, opened 2020-06-23
Operator definition with PickOperator implementation

van-fs:
This PR introduces the Operator artifact.
- It differs from the functional approach in the prototype. `Operator` defines the abstract class that an instance like `PickOperator` will implement.
- There's an enum for `OperatorValidationError`.  I thought that if Operators were reporting validation issues themselves, we might see a lot of the same strings used, which would bloat code.
- `PickOperators` tests were beefed up for the new validation feature and some additional cases.


src/operator.ts
```diff
@@ -0,0 +1,65 @@
...
+ * return a result that will be passed to the next operator, and it should not mutate the incoming object.
+ *
+ * While an Operator will most often transform a single object, the input is always a list of objects.
+ * This is because data emitted from the data layer could be a single value or the result of a function call, which
+ * emits a list of arguments.
+ *
+ * An Operator can choose not to pass information to the next operator by returning null.
+ */
+export abstract class Operator {
+  readonly name: string;
```
TrevorFSmith, line 31: Would it make sense to hold the OperatorOptions instance instead of transferring its data to properties on Operator? This would give us some wiggle room for additions to options (both data and logic) in the future without changing the constructor API.

src/operators/pick.ts
```diff
@@ -0,0 +1,74 @@
+import { Operator, OperatorOptions, OperatorValidationError } from '../operator';
+
+export interface PickOperatorOptions extends OperatorOptions {
+  properties: string[] | string;  // the properties in the object to pick as a list or CSV
+}
+
+type picktionary = { [key: string]: boolean };
```
TrevorFSmith, line 7: hee hee

src/operator.ts
van-fs: Made changes to accommodate `options` as a member of the class.  Due to TS compiler checking, the `index` member needs to remain as it must hold a value to support `data[index]` usage in `handleData()`.  Trevor and I both agreed on the approach.

PR #4, opened 2020-06-24
Add logging utility to send events to FullStory

van-fs:
This PR adds a logging utility designed to send Custom Events to FullStory.  It provides health status for DLO after deployment.
- It's not a generic logging utility, and I've left out the concept of logging levels. I had this in the prototype and found that I didn't really use error versus warn.  It also just makes for more choices that I'd rather limit right now.
- I've included common "classes" of issues in `LogEventTypes` enum. These will be used as we build out the implementation.
- If you want details like an `Error.message` it can be used in `send`.
- Additionally, a `datalayerPath` will help identify the target data layer.  Often this was needed to know which data layer was not on page due to timing issues.


van-fs:
One topic for consideration, we could leverage `@fullstory/browser` to also provide API types.  But I believe there's some work that would need to be done to define a top level interface to do so.  At present `@fullstory/browser` exports a function signature for `FS.event` but not a `FS` interface that would contain `event`.

van-fs:
Excellent suggestion; I will refactor today.

patrick-fs:
> Excellent suggestion; I will refactor today.

Agreed. One other consideration are the quota limits on `FS.event` - 30/minute, 10/second burst per session. Depending on how chatty the logging is, customer quotas could be exhausted if this were configured in production. Any `FS.event` instrumentation could be lost.

Is the primary use-case to configure logging in prod or is there a use-case where a developer registers a logger *only for their session* in the browser console?

src/utils/logger.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export enum LogEventTypes {
+  DATALAYER_NOT_FOUND = 'Data Layer Not Found',
+  EMIT_FAILURE = 'Emit Failure',
+  OPERATOR_FAILURE = 'Operator Failure',
+  LISTENER_FAILURE = 'Listener Failure',
+  RULE_FAILURE = 'Rule Failure',
+  INCORRECT_ARGUMENTS = 'Incorrect Arguments',
+  UNKNOWN_INPUT = 'Unknown Input',
+}
```
patrick-fs, line 11: Docs TODO: The conditions that trigger these log event types should be enumerated.

van-fs:
See the updated design.  The FullStory specific calls were pulled out of the `Logger` utility.  It now supports a generic way to send `error`, `warn`, `info`, and `debug` messages.  It's up to the `LogAppender` to decide how to handle serializing these.  A FullStory custom event serializer exists in the test.  Higher layers will need to decide when/how to register this. Perhaps this is what resides in the snippet: an expando that is a function to emit custom events using FS.

To answer @patrick-fs question, the primary use case is monitoring the health of a DLO deployment - not as an always on logging utility.  In prototyping, I mainly (and sparingly) used error conditions to send events.  There is a `level` setting that could govern the chativity of the Logger, but the more logging that goes into the core, the larger the package size will be.  This is why I was first using `LogEventTypes` as generic classes, but let's see if these can be defined closer to where the logging is called.

van-fs:
Once the MockClass PR merges I'll update this PR to use it. That will cut down on some of the code in the test suite.

van-fs:
@TrevorFSmith I made some modifications to use your MockClass.  I had to fix a few typescript compiler complaints, and I'll get a tsconfig.json into another PR.

The MockClass worked really well, and I added two convenience `expect` functions.  I found myself having to use the non-null assertion so I created an expect that checks the `callQueue` length is one and then returns a non-null value.  I also added an `expectParams` to go one step farther and it makes destructuring params more convenient:

```
    FS.log(0, 'value')
    const [ logLevel, message ] = expectParams(FS, 'log');
    expect(logLevel).to.eq(0);
    expect(message).to.eq('value');
```

TrevorFSmith:
Ok, reviewing now...

src/utils/window.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Defines a subset of FullStory browser APIs used by DLO.
+ */
+export interface FullStoryAPI {
```
TrevorFSmith, line 4: Is this vestigial now that we have the FS mock and the logger uses the listener pattern?

test/mock.ts
```diff
@@ -21,5 +23,27 @@ export class MockClass {
...
+  constructor(public parameters: any[], public result: string | null | void){}
+}
+
+/**
+ * Tests whether a call queue has one Call and returns it.
+ * @param mock the MockClass
+ * @param methodName the call queue's method name
+ */
+export function expectCall(mock: MockClass, methodName: string): Call {
+  expect(mock.callQueues[methodName].length).to.eq(1);
```
TrevorFSmith, line 35: Could this test be to expect that the length is greater than zero or to pass an optional parameter to set the expected length, with -1 meaning don't check? That way we can use expectCall in different situations other than a clean, one-call test.

src/utils/window.ts
van-fs: Good catch; yes.

test/mock.ts
van-fs: Providing the -1 no check option is difficult since the expectCall was intended to always return a value.  So there's an optional now if there is an expected length; else, it assume it's more than 0.  If you want to check that there were truly no calls, I added an `expectNoCalls` function.

PR #5, opened 2020-06-26
Added interfaces for the full CEDDL spec

TrevorFSmith:
This implements the various interfaces for CEDDL and creates a couple of instances of it for use during testing.

@van-fs One question: I wrote a test for CEDDL using the pick operator and it is failing. So, I've added a pick operator test that (I think) reproduces the failure in the context of the pick tests. Am I missing something about how the properties option is able to specify that a property at any depth can be picked?

### DIFF
--- a/test/ceddl.spec.ts
+++ b/test/ceddl.spec.ts
@@ -2,14 +2,40 @@ import { expect } from 'chai';
 import 'mocha';
 
 import * as CEDDL from './data/CEDDL';
+import { PickOperator } from '../src/operators';
 
 describe('mock CEDDL unit tests', () => {
 
-  it('it should have JSO subobjects ', () => {
-    const { minimalDigitalData } = CEDDL;
-    const { pageInstanceID, page, product, cart, transaction, event, component, user, privacy, version } = minimalDigitalData;
+  it('empty digitalData should have JSO subobjects ', () => {
+    const { emptyDigitalData } = CEDDL;
+    const {
+        pageInstanceID, page, product, cart, transaction,
+        event, component, user, privacy, version
+    } = emptyDigitalData;
 
-    expect(minimalDigitalData).not.be.undefined;
+    expect(emptyDigitalData).not.be.undefined;
+    expect(pageInstanceID).not.be.undefined;
+    expect(pageInstanceID).be.empty;
+    expect(page).not.be.undefined;
+    expect(product).not.be.undefined;
+    expect(cart).not.be.undefined;
+    expect(transaction).not.be.undefined;
+    expect(event).not.be.undefined;
+    expect(component).not.be.undefined;
+    expect(user).not.be.undefined;
+    expect(privacy).not.be.undefined;
+    expect(version).not.be.undefined;
+    expect(version).not.be.empty;
+  });
+
+  it('basic digitalData should have JSO subobjects ', () => {
+    const { basicDigitalData } = CEDDL;
+    const {
+        pageInstanceID, page, product, cart, transaction,
+        event, component, user, privacy, version
+    } = basicDigitalData;
+
+    expect(basicDigitalData).not.be.undefined;
     expect(pageInstanceID).not.be.undefined;
     expect(pageInstanceID).not.be.empty;
     expect(page).not.be.undefined;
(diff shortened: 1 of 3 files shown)


test/data/CEDDL.ts
```diff
@@ -23,8 +25,10 @@ export interface CEDDL {
 /**
  * Because of the wide range of methods for categorization, an object literal is provided for categories.
  */
-export interface Category {
+export interface PageCategory {
```
van-fs, line 28: I had noticed that there was some overlap in the *Category models and had created a generic `Category` object.  But I think what you've done is encapsulate those shared properties in each domain category - presumably because changing them would not depend on updating the super interface.  I think that this is the right approach.

test/ceddl.spec.ts
```diff
@@ -24,4 +50,21 @@ describe('mock CEDDL unit tests', () => {
     expect(version).not.be.empty;
   });
 
+
+  it('it should pick data from CEDDL', () => {
```
van-fs, line 54: Should we:
A) Reference CEDDL test data in any and all operator's tests
B) Only use mock data defined within the scope of each tests

During prototyping there was feedback that this caused the developer to step out of the test suite to review external test data.  Given we're supporting a specific spec, perhaps we need to simply reference.  This would cut down on the amount of arbitrary test data and provide an artifact for all operator developers.

test/ceddl.spec.ts
```diff
@@ -24,4 +50,21 @@ describe('mock CEDDL unit tests', () => {
...
+
+    const pick = new PickOperator({
+      name: 'pick',
+      properties: 'version,pageID'
+    })
+    const output = pick.handleData([basicDigitalData]);
+    console.log('output', output);
+    expect(output).to.not.be.null;
+    expect(output![0].version).to.eq(ceddlVersion);
+    expect(output![0].pageID).to.eq(basicDigitalData.page.pageInfo.pageID);
```
van-fs, line 67: I may need to clarify/revisit the design of `pick` based on this test.  This test fails because `pageID` won't be found. The reason is because neither `page` or `pageInfo` is provided as a picked property.  So as the picker traverses into child properties, it will encounter `page` and stop traversal because `page` is not listed as a picked property.  This design was intended to guard against picking a property with the same name but in different children.  By providing the parent properties you can better manage the traversal and picked properties.

test/operator-pick.spec.ts
```diff
@@ -87,6 +87,15 @@ describe('pick operator unit tests', () => {
     expect(output![0].contact.email).to.eq(data.contact.email);
   });
 
+  it('it should find properties at different depths', () => {
```
van-fs, line 90: Similar needs: decide on whether to use CEDDL mock or local test data, and add `favorites` to pick list.

test/ceddl.spec.ts
TrevorFSmith: In general, I think it makes sense to test the operators in their own spec files and just test CEDDL itself in here. This should move into operator-pick.spec.ts.

test/ceddl.spec.ts
TrevorFSmith: My initial thought was that that pick operator would take dot-path notation like selectors, so the `properties` value would be `version,page.pageInfo.pageID`. The tests made it clear that wasn't the case but perhaps it's something to consider?

test/ceddl.spec.ts
van-fs: Yeah, I think we need a more expressive but lightweight syntax: needs to support dot notation, an array index, and maybe a * wildcard - on properties names and `parent.*`.  Defining the grammar and parser outside would get away from these idiosyncracies.

test/ceddl.spec.ts
van-fs: Yeah, I think we need a more expressive but lightweight syntax: needs to support dot notation, an array index, and maybe a * wildcard - on properties names and `parent.*`.  Defining the grammar and parser outside would get away from these idiosyncracies.

TrevorFSmith: Agreed. Since we need the same syntax for rule source selection it seems like building a separate parser/data-selector method makes sense.

test/ceddl.spec.ts
TrevorFSmith: I now see your original question. I think it's fine to test against CEDDL instead of a test-local data structure since it's a standard data layer object that I would guess that most readers will know.

test/operator-pick.spec.ts
```diff
@@ -87,6 +88,15 @@ describe('pick operator unit tests', () => {
...
+    const pick = new PickOperator({ name: 'pick', properties: 'id,color' })
```
patrick-fs, line 92: suggestion: an array of strings as the value for `properties` may be more interpretable for API users rather than a delimited string

van-fs: It's a good suggestion.  PickOperatorOptions supports a string or an array of strings.  There are two tests for this in the suite:
- it should pick array properties from an object
- it should pick csv properties from an object

patrick-fs: Got it - I see it now.

PR #6, opened 2020-06-29
Mock up a FullStory recording API for use in unit testing

TrevorFSmith:
Putting this up for comment by @van-fs before going much farther with it. Specifically, it needs a review of the mocking method.

After looking at current mock libs like Sinon (complex) and our needs for mocking FS's recording APIs (simple) I whipped up a basic `MockClass` that (when extended) records calls to any methods on a class for checking during a test.

Take a look at `test/vendor/fullstory-recording.ts` and `test/vendor.spec.ts` to see how we could test that the FS recording API is called (eventually) by the DLO.

### DIFF
--- a/test/vendor/fullstory-recording.ts
+++ b/test/vendor/fullstory-recording.ts
@@ -1,0 +1,33 @@
+import { MockClass } from '../mock';
+
+export class FullStory extends MockClass {
+  consent(userConsents?: boolean): void {}
+
+  disableConsole(): void {}
+  
+  enableConsole(): void {}
+  
+  event(eventName: string, eventProperties: object): string | null {
+    return null
+  }
+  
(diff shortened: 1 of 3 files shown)


test/vendor/fullstory-recording.ts
```diff
@@ -0,0 +1,33 @@
+import { MockClass } from '../mock';
+
+export class FullStory extends MockClass {
+  consent(userConsents?: boolean): void {}
+
+  disableConsole(): void {}
+  
+  enableConsole(): void {}
+  
+  event(eventName: string, eventProperties: object): string | null {
```
van-fs, line 10: Can we add source as an optional 3rd argument? We'll be using it to track usage.

PR #8, opened 2020-07-01
Path selection utility methods.

TrevorFSmith:
This will eventually provide the path selection utility methods used to find rule sources and within operators. The goal is to use path selection strings like `digitalData.page.pageInfo.author` or `digitalData.cart.item[-1].quantity` to traverse the Javascript namespace and find specific information.

### DIFF
--- a/src/selector.ts
+++ b/src/selector.ts
@@ -60,7 +60,7 @@ For example: `!(foo=bar, blatz)` where `!` is the kind and `foo=bar` and `blatz`
 */
 class Op {
   kind: OpKind;
-  index?: Number;
+  index: number = 0;
   props: OpProp[] = [];
 
   constructor(public raw: string) {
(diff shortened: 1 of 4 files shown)


TrevorFSmith:
I've implemented pluck `foo.bar.blatz` and index `foo[-1]` selection so it seems like a good time to merge so that Van can write against this.

Everything parses, I just haven't written the selection code for the remaining operations. It will throw a TBD error if it hits those paths.

src/selector.ts
```diff
@@ -0,0 +1,298 @@
...
+    if (index >= prop.length) return undefined;
+    if (index < 0) {
+      index = prop.length + index
+    }
+    if (index < 0) return undefined;
+
+    try {
+      return prop[index];
+    } catch (e) {
+      console.log('Index fail', this.brackets, e);
```
van-fs, line 210: In the future we might remove console.logs (might need to get a linter to assist).

TrevorFSmith: Yeah, I generally remove them but missed this one.

PR #13, opened 2020-07-02
Add DataHandler to process event data

van-fs:
This PR introduces the `DataHandler` to respond to and process event data using operators:
- Events emitted from future `PropertyListener`s will emit `CustomEvent`s, which will use `DataLayerDetail` to provide metadata from the data layer. Individual `FunctionDetail` and `PropertyDetail` classes further disambiguate the type of event from the data layer.
- The `DataHandler` responds to events and processes based on the data layer `path`.  The `path` is used by the handler to decide if the event data should be processed: event listeners all listen to the same `type` event so something must be used to choose when to run the handler.  The multicast design is provided by the browser - it's not ours.
- There's a `debug` flag specifically in the handler instance. This allows the user to debug transformations for a given handler rather than sifting through gobs of log data.
- The `debugger` is a simple function but one that could be overridden by tooling (i.e. WYSIWG builder).
- Note the `object` utility will be removed in favor of the query syntax.

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -1,0 +1,25 @@
+export enum DataLayerEventType {
+  PROPERTY = '_fs_DataLayerProperty',
+  FUNCTION = '_fs_DataLayerFunction',
+}
+
+export class DataLayerDetail {
+  value?: object;
+  args?: any[];
+
+  constructor(readonly target: object, readonly path: string) {
+
+  }
+}
+
+export class FunctionDetail extends DataLayerDetail {
+  constructor(target: object, public args: any[], path: string) {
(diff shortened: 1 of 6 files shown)


src/event.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Defines CustomEvent types.
+ * See https://developer.mozilla.org/en-US/docs/Web/API/Event/type
+ */
+export enum DataLayerEventType {
+  PROPERTY = '_fs_DataLayerProperty',
```
TrevorFSmith, line 6: Can we change these to `_dlo_*` ?

PR #18, opened 2020-07-06
adding eslint + git pre-commit hook

patrick-fs:
This PR:

- Installs eslint with the airbnb style guide and adds two new npm scripts: `lint` which executes eslint in the `src` and `test` directories and `lint:fix` which does the same but adds the `--fix` argument to automatically fix any linting errors that are automatically fixable
- Installs the `pre-commit` package and executes `lint:fix` in a git pre-commit hook

Right now there are 93 eslint issues that can't be automatically fixed (and about 120 that can). These 93 errors must be fixed manually. 

With the pre-commit hook in place, you'll need use the `-n` (no-verify) argument when committing files while working through the linting issues:

```
git add {whatever}
git commit -n -m "some message"
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,19 @@
+{
+    "env": {
+        "browser": true,
+        "es2020": true
+    },
+    "extends": [
+        "airbnb-base"
+    ],
+    "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "ecmaVersion": 11,
+        "sourceType": "module"
+    },
+    "plugins": [
+        "@typescript-eslint"
+    ],
+    "rules": {
+    }
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -27,5 +27,13 @@
     "import/extensions": ["error", "ignorePackages", {
       "ts": "never"
     }]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["test/**/*.spec.ts"],
+      "rules": {
+        "no-unused-expressions": "off"
+      }
+    }
+  ]
 }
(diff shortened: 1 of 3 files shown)


patrick-fs:
there are currently 151 linting problems that can't be fixed automatically

.eslintrc.json
```diff
@@ -0,0 +1,39 @@
+{
+  "env": {
+    "browser": true,
+    "es2020": true,
```
patrick-fs, line 4: currently set to accept es2020 (ecmaVersion 11) syntax. This is probably fine, since the TypeScript compiler will compile down to the target ES version appropriate for browsers.

.eslintrc.json
```diff
@@ -0,0 +1,39 @@
...
+    "mocha": true
+  },
+  "extends": [
+    "airbnb-base",
+    "plugin:import/typescript"
```
van-fs, line 9: Overall LGTM.  I noticed one lint for example:

```
fullstory-data-layer-observer/test/utils/mocha.ts
   3:10  error  'MockClass' is defined but never used                                  no-unused-vars
   3:21  error  'Call' is defined but never used                                       no-unused-vars
```

That seems odd since the import is only providing type safety.  Given this [discussion on SO](https://stackoverflow.com/questions/57802057/eslint-configuring-no-unused-vars-for-typescript), looks like there's a recommended rule set.  But rather than take it en massed, it can be scoped down to a specific rule:

```
{
      "files": ["*.ts"],
      "rules": {
        "no-unused-vars": "off",
        "@typescript-eslint/no-unused-vars": ["error"]
      }
    }
```

I'll include a linted PR with the above and fixes.

PR #21, opened 2020-07-07
add linting to ci build

patrick-fs:


.circleci/config.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       - run:
           name: Audit
           command: npm audit --audit-level=low --production
+      - run:
```
van-fs, line 25: I was actually wondering if the lint step would be a required part of the build.  Like you can't build unless linting passes.  So there'd be an `npm build: '<lint command> <build command>'`.  If it makes sense to get linting in sooner than later, LGTM.

patrick-fs: Yeah, that's what we'd do. Linting and testing happen before the build command.

PR #22, opened 2020-07-08
Resolve lint errors

van-fs:
This PR addresses lint errors that were overlooked on the first pass.
- The package.json's pattern match for eslint command was updated
- Changes are mostly superficial
- @TrevorFSmith can you do a 3rd pass and review selector and submit PR?


TrevorFSmith:
👁👁

TrevorFSmith:
Ok, here are the lint errors in selector.js that I don't believe we should fix:
image

`no-continue` is misguided. short circuiting loops is so much better than just wrapping all of the logic in an `if` statement.

'prefer-destructuring` is also misguided, especially since the syntax is terribly awkward when assigning values to class properties.

On line 319 I explicitly want to use loose equality and don't really know of a way to work around that to make `eqeqeq` happy.

With modern editors that can toggle between soft and hard wrap, do we really want to enforce `max-len`? If so, 80 makes more sense than 100.

patrick-fs:
> Ok, here are the lint errors in selector.js that I don't believe we should fix:

> `no-continue` is misguided. short circuiting loops is so much better than just wrapping all of the logic in an `if` statement.

`no-continue` is definitely a style opinion. Per the eslint docs: _When used incorrectly it makes code less testable, less readable and less maintainable._

I see the merits of this argument but personally don't have a strong opinion - if you want to create an override in the eslint config, I'm fine with that.

> 
> `prefer-destructuring` is also misguided, especially since the syntax is terribly awkward when assigning values to class properties.

Go ahead and turn off array destructuring for the file. Adding this comment at the top should do the trick:
`/* eslint prefer-destructuring: ["error", {AssignmentExpression: {array: true}}] */`

> On line 319 I explicitly want to use loose equality and don't really know of a way to work around that to make `eqeqeq` happy.

Can you please add additional context in a comment about why the value must loosely match and  then a single eslint ignore comment that applies only to that line of code.

> With modern editors that can toggle between soft and hard wrap, do we really want to enforce `max-len`? If so, 80 makes more sense than 100.

I think this is also inclusive of folks who are viewing the file on GitHub and not in an IDE. A max length should prevent horizontal scrolling. If you feel strongly about changing the length from 100 to 80, feel free to do so. It should be a global change.

TrevorFSmith:
The GitHub argument for `max-length` sounds reasonable. The width of GitHub text areas is around 140 chars so how about we set the `max-length` to that?

van-fs:
Do I have a `141` ... going once, twice, `140` it is.

package.json
```diff
@@ -44,4 +46,4 @@
     "ts-node": "^8.10.2",
     "typescript": "^3.9.5"
   }
-}
+}
```
patrick-fs, line 49:
```suggestion
}

```

PR #23, opened 2020-07-08
Adding DataLayerObserver core

van-fs:
This PR adds the main functionality of DataLayerObserver for static data layers.  The `readOnLoad` property is heavily used in this iteration.  The `FunctionOperator` is a key component since it is used with the `destination` in rules.

PR #24, opened 2020-07-08
refactoring Operator class - removing Generic implementation

patrick-fs:
Hey @van-fs I got a path towards the type-checking you are looking for. Check this out.

### DIFF
--- a/src/operators/function.ts
+++ b/src/operators/function.ts
@@ -3,6 +3,7 @@ import { select } from '../selector';
 
 export interface FunctionOperatorOptions extends OperatorOptions {
   func: string | Function;
+  thisArg?: string | object;
 }
 
 /**
@@ -10,13 +11,28 @@ export interface FunctionOperatorOptions extends OperatorOptions {
  */
 export class FunctionOperator extends Operator<FunctionOperatorOptions> {
   handleData(data: any[]): any[] | null {
-    const { func } = this.options;
+    const { func, thisArg } = this.options;
+
+    let actualThisArg: object = globalThis;
+
+    if (thisArg) {
+      switch (typeof thisArg) {
+        case 'object':
+          actualThisArg = thisArg;
+          break;
+        case 'string':
+          actualThisArg = select(thisArg);
+          break;
+        default:
+          super.throwValidationError('thisArg', OperatorValidationError.UNSUPPORTED);
+      }
+    }
 
     switch (typeof func) {
       case 'function':
-        return [func.apply(globalThis, data)];
+        return [func.apply(actualThisArg, data)];
       case 'string':
-        return select(func).apply(globalThis, data);
+        return select(func).apply(actualThisArg, data);
       default:
         // NOTE this will stop the handler
         return null;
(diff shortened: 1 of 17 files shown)


src/operators/function.ts
```diff
@@ -9,9 +9,14 @@ export interface FunctionOperatorOptions extends OperatorOptions {
 /**
  * FunctionOperator executes a function and returns the result.
  */
-export class FunctionOperator extends Operator<FunctionOperatorOptions> {
+export class FunctionOperator extends Operator {
```
patrick-fs, line 12: Removing the generic type variable on the `Operator` class impacted this class the most. You can see how I have to include a "useless constructor" as well as explicitly cast the `options` instance when it is referenced.

test/observer.spec.ts
```diff
@@ -147,7 +146,7 @@ describe('DataLayerObserver unit tests', () => {
     });
 
     expect(observer.handlers.length).to.eq(1);
-    // @ts-ignore TODO (van) how to typecheck this
+
```
patrick-fs, line 149: tada, `ts-ignore` statements can be removed

patrick-fs: I think want you really want is 

```
registerOperator(name: string, operator: InstanceType<typeof Operator>) {
```

but in order to do that, you'll need to make the `Operator` non-abstract. That will require a little more gymnastics.

van-fs: @patrick-fs Thank you.  Looking at it now.  I think the dilemma that exists is that it may be difficult to enforce the type safety for module developers AND have "generics" to support a DSL using DataLayerRule.  Might have to pick the one that has less consequences for code maintenance, and that at present looks like moving to this type of a re-design.  Fortunately @TrevorFSmith had suggested a `validate` function so even if we move to generic options that we don't know the shape of, we can always use validate to ensure the options are semantically correct.  I believe there might be additional convenience functions needed to help validate rules easily so it's not a gnarly bunch of if/else blocks.  @TrevorFSmith, do you want to approve the existing #23 in isolation and incorporate these changes in another PR or simply make them now in #23 ?

TrevorFSmith: I'd prefer to handle them separately so I'll go review #23 now.

TrevorFSmith:
This now has conflicts with `main`. @patrick-fs and @van-fs are you of the 'merge main and FF' or the 'rebase from main' inclinations?

patrick-fs:
> This now has conflicts with `main`. @patrick-fs and @van-fs are you of the 'merge main and FF' or the 'rebase from main' inclinations?

There are oodles of conflicts. I usually prefer rebase from main, but it might be more straight-forward to redo the changes that I mad on a new branch (I really didn't do much).

patrick-fs:
> This now has conflicts with `main`. @patrick-fs and @van-fs are you of the 'merge main and FF' or the 'rebase from main' inclinations?

I usually prefer rebase from main, but it looks like it might be tough to handle in this case. merge and FF might be better.

patrick-fs:
Closing because https://github.com/fullstorydev/fullstory-data-layer-observer/pull/25

PR #25, opened 2020-07-10
Refactor to support interfaces versus generics

van-fs:
I took @patrick-fs PR and riffed on it.  I've moved the `Operator` away from an abstract class to a basic interface.

```
export interface Operator {
  options: OperatorOptions;
  handleData(data: any[]): any[] | null;
  validate(): void;
}
```

To make it so that `DataLayerObserver` can deal with arbitrary metadata coming from `DataLayerRule`, an `OperatorFactory` was created.  The factory internalizes the creation of an `Operator` to it versus doing it in the `DataLayerObserver`.  The factory will make for a cleaner separation and way to organize operators we provide.  To support some of the type checking, we'll have to maintain two union typedefs as we build more operators. 

```
export type BuiltinOperator = typeof FunctionOperator | typeof SomeFutureOperator;
export type BuiltinOptions = FunctionOperatorOptions | typeof SomeFutureOperatorOptions;
```

Since a `DataLayerRule` is given to us by users and could contain errors, I've created a more robust `OperatorValidator` for runtime validation. This has a few convenience methods but the main one is `validate`, which accepts an `OperatorSpecification`.  This gives us a JSONy way to declare the expected options at runtime and prevents ugly if/else blocks.  Operator implementors can then just use:

```
const validator = new OperatorValidator(this.options);
validator.validate(FunctionOperator.specification);
```

The `specification` itself is internal to the `Operator`.  I don't think there's a need to externalize it, but we could.  It's also static since it's shared across instances.  The `Operator` interface doesn't force you to supply a specification, but it could.  The main rationale being that validation is possibly more than a specification and `Operator` implementors are free to roll their own `validate` method.

All current tests are passing.  I'll create another spec to cover the validator features to bump up the coverage if we approve this.

### DIFF
--- a/test/observer.spec.ts
+++ b/test/observer.spec.ts
@@ -103,10 +103,6 @@ describe('DataLayerObserver unit tests', () => {
     expectNoCalls(globalMock.console, 'log');
   });
 
-  it('a missing data layer should be retried and fail gracefully', () => {
-
-  });
-
   it('it should allow custom operators to be registered', () => {
     const observer = new DataLayerObserver();
 
@@ -176,10 +172,6 @@ describe('DataLayerObserver unit tests', () => {
     expect(cart).to.eq(globalMock.digitalData.cart);
 
     expectNoCalls(globalMock.console, 'log');
-  });
-
-  it('rule validation can be enabled', () => {
-
   });
 
   it('rules can be previewed before making them live', () => {
--- a/test/observer.spec.ts
+++ b/test/observer.spec.ts
@@ -3,20 +3,17 @@ import { expect } from 'chai';
 import 'mocha';
 
 import { DataLayerObserver } from '../src/observer';
-import { basicDigitalData, CEDDL } from './mocks/CEDDL';
+import { basicDigitalData, CEDDL, PageCategory } from './mocks/CEDDL';
 import Console from './mocks/console';
 import FullStory from './mocks/fullstory-recording';
 import { expectParams, expectNoCalls } from './utils/mocha';
 import { Operator, OperatorOptions } from '../src/operator';
 
 class EchoOperator implements Operator {
-  options: OperatorOptions;
-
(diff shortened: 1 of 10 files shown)


van-fs:
One other notable design point.  Users invoke `registerOperator` with simply an instance of an `Operator`.  This is a bit different than how we build operators using the factory.  It's really hard to generalize registration of 3rd party operators and get the TS compiler to not complain.  So 3rd parties will register their pre-configured operators and then just reference them by name when they create a `DataLayerRule`.  (The main difference being is that they can't supply `OperatorOptions` for registered operators the same way they can for our built-in operators.)

test/observer.spec.ts
```diff
@@ -7,21 +7,25 @@ import { basicDigitalData, CEDDL } from './mocks/CEDDL';
...
-import { FunctionOperator } from '../src/operators';
+import { Operator, OperatorOptions } from '../src/operator';
 
-class EchoOperatorOptions implements OperatorOptions {
-  name = 'echo';
-}
+class EchoOperator implements Operator {
+  options: OperatorOptions;
+
+  constructor(options?: OperatorOptions) {
```
patrick-fs, line 15: I think you can collapse this into a default value:

```
constructor(options?: OperatorOptions = { name: 'echo' }) {
...
```

test/observer.spec.ts
```diff
@@ -7,21 +7,25 @@ import { basicDigitalData, CEDDL } from './mocks/CEDDL';
...
+    this.options = options || {
+      name: 'echo',
+    };
```
patrick-fs, line 18: I think you can collapse this into a default value:

```
constructor(options?: OperatorOptions = { name: 'echo' }) {
...
```

van-fs:
@TrevorFSmith Last round of changes have been made.  It's a bigger PR than I'd like so let me know if you need any verbal overview.

PR #26, opened 2020-07-10
Build the embed script and use a snippet

TrevorFSmith:
One way to use the DLO is to embed it into a page as a bundled script and use in-page variables to configure it.

This PR will start the process of building, bundling, and then initializing using a "snippet" of HTML.


TrevorFSmith:
@van-fs and @patrick-fs
After evaluating webpack and rollup I'm going with rollup since it is pretty much the same in terms of output and the configuration is nicer.

For the snippet, I've created a small example, `/examples/basic-embed/index.html`, of how someone would run the DLO using a self-hosted build and with snippet variables. I started with Van's POC template and tweaked it based on later conversations. I'd like you both to weigh in before I write the initialization logic that uses the snippet, which will live in `/src/embed/index.ts` 

In a future PR we can figure out CDN hosting and whether we need the same script element creation code as the recording script.

TrevorFSmith:
@van-fs Yes, the bundled script will execute whatever is in `init.ts` so we can put all of the snippet logic like parsing vars on `window` in there.

Regarding the name, `embed`: It's not a rollup convention but I chose it because it indicates that the bundle is meant to be embedded in a browser instead of put into a "prebuilt" npm package (like our React Native babel plugin is right now). I feel like `script` is somewhat more vague since modules also include scripts.

src/embed/init.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // Read rules
+  const rules = _dlo_collectRules();
+  if (rules.length === 0) {
+    /* eslint-disable no-console */
+    console.warn('No rules for the Data Layer Observer');
+  }
+
+  win._dlo_observer = new DataLayerObserver({
+    previewMode: preview,
```
van-fs, line 70: Thoughts on initializing `previewMode` here using something like `!!win._dlo_preview` or a ternary.  Any need to have a separate var and if check?  The same for `readOnLoad`.

src/embed/init.ts
```diff
@@ -0,0 +1,76 @@
...
+
+function _dlo_collectRules(): any[] {
+  const results: any[] = [];
+  Object.getOwnPropertyNames(window).forEach((propName) => {
+    if (propName.startsWith('_dlo_rules') === false) return;
+
+    const prop = (window as { [key: string]: any })[propName];
+    if (Array.isArray(prop) === false) {
+      /* eslint-disable no-console */
+      console.warn(`window[${propName}] is not an array so will be ignored`);
```
van-fs, line 38: Can we use `Logger.getInstance().warn('No rules for the Data Layer Observer')`?  That will by default use the `ConsoleAppender`, but I think what we also want to do is set up FullStory as the log appender.  The `logger.spec.ts` has a unit test to add FullStory:

```
const logger = Logger.getInstance();
    logger.appender = {
      log(event: LogEvent) {
        // eslint-disable-next-line camelcase
        const { level: level_int, message: message_str, datalayer: datalayer_str } = event;
        FS.event('Data Layer Observer', { level_int, message_str, datalayer_str }, 'dataLayerObserver');
      },
    };
```

Can we do something similar by creating a `_dlo_log` property or similar to assign a custom function.  Our default snippet would then provide the correct implementation to use FullStory.

src/embed/init.ts
```diff
@@ -0,0 +1,76 @@
...
+  let readOnLoad = false;
+  if (win._dlo_readOnLoad === true) {
+    readOnLoad = true;
+  }
+
+  // Read rules
+  const rules = _dlo_collectRules();
+  if (rules.length === 0) {
+    /* eslint-disable no-console */
+    console.warn('No rules for the Data Layer Observer');
```
van-fs, line 66: Another possible `Logger` candidate.

src/embed/init.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  win._dlo_observer = new DataLayerObserver({
```
van-fs, line 69: We probably need to also have a conditional check on `win._dlo_observer`.  If a domain team acted independently and included `dlo.js` init would try to recreate a `DataLayerObserver`, right?

We could check `win._dlo_observer` and if set, silently stop processing.  But that means a team who dynamically loaded DLO via a tag manager would not have their rules processed.  So there needs to be some incremental processing done for newly found rules, which means we probably need a list of previously processed  _dlo_rules* expandos.  Within init we could check for `win._dlo_observer` and if set, it means that we should look for new rules and then use `win._dlo_observer.processRule(rule) on the incremental ones.`. If needed, `processRule` could be updated to accept a list as convenience.

van-fs: More thoughts.  Previously the pattern was a singleton.  That presents difficulty if one team wants to preview their contributed rule sets (`previewMode` is a global config setting).  We could create a `DataLayerObserver` object for each team -  allowing true separation of rules and config, but that's probably a good amount of design work to itemize and validate the cases.  We can call out in the documentation if only one `DataLayerConfig` rules them all.

src/embed/init.ts
van-fs: There are a few other properties that should be read from window expando per the `DataLayerConfig`:

```
/**
 * DataLayerConfig provides global settings for a DataLayerObserver.
 *
 * Required
 *  rules: a list of pre-configured DataLayerRules
 * Optional
 *  beforeDestination: OperatorOptions that is always used just before before the destination
 *  previewMode: redirects output from a destination to previewDestination when testing rules
 *  previewDestination: output destination using selection syntax for with previewMode
 *  readOnLoad: when true reads data layer target(s) and emit the initial value(s)
 *  validateRules: when true validates rules to prevent processing invalid options
 *  urlValidator: a function used to validate a DataLayerRule's `url`
 */
export interface DataLayerConfig {
  beforeDestination?: OperatorOptions;
  previewDestination?: string;
  previewMode?: boolean;
  readOnLoad?: boolean;
  rules: DataLayerRule[];
  validateRules?: boolean;
  urlValidator?: (url: string | undefined) => boolean;
}
```

src/embed/init.ts
TrevorFSmith: I've added a check for duplicate assignment of window._dlo_observer that emits an error and bails so we won't create more than one.
For the sake of clarity and simplicity of deployment, what do you think about using the singleton pattern for embed+snippet and if customers want more complex setups then they can directly use the modules?

van-fs: @TrevorFSmith I like the simple approach.  We expect one DLO and corresponding config.  If you load more than one, we won't break but it's not going to do anything.  We will support rules contributed from other teams.  I think that's where we are currently, right?

examples/basic-embed/index.html
```diff
@@ -0,0 +1,78 @@
...
+      // Default is null
+      window['_dlo_previewDestination'] = 'console.log';
+
+      // When true reads data layer target(s) and emits the initial value(s)
+      // Default is false
+      window['_dlo_readOnLoad'] = true;
+
+      // When true validates rules to prevent processing invalid options
+      // Default is false
+      window['_dlo_validateRules'] = false;
```
van-fs, line 57: I'm thinking since `previewMode` is set to true that we also set `_dlo_validateRules` set to true.  So the example is in "safe mode".

examples/basic-embed/index.html
```diff
@@ -0,0 +1,78 @@
...
+      // Default is null
+      window['_dlo_beforeDestination'] = null;
+
+      // Redirects output from a destination to previewDestination when testing rules
+      // Default is false
+      window['_dlo_previewMode'] = true;
+
+      // The output destination using rule selection syntax for use with previewMode
+      // Default is null
+      window['_dlo_previewDestination'] = 'console.log';
```
van-fs, line 49: One point of clarification that I could use perspective on: under the hood, DLO will default to `console.log` if `_dlo_previewDestination` is falsy.  So is `Default is null` the right wording?

src/embed/init.ts
TrevorFSmith: Yep, the DLO will look for anything on `window` that starts with `_dlo_rule` so different teams can add their own array or rules and it will be picked up.

examples/basic-embed/index.html
TrevorFSmith: Hmmm. If I set previewMode true and don't pass in a previewDestination then I don't see console logs. It silently drops the output.

PR #30, opened 2020-07-13
FullStory CEDDL user compatibility

van-fs:
This PR introduces CEDDL user compatibility to FullStory.  Since this is the first CEDDL-to-FullStory ruleset, there's some upfront design work:
- I created a `examples/rules` directory.  Trevor already has an `examples` folder so this would continue that direction.
- Each file in the `examples/rules` would have a <vendor>.ts file with exported rules.  The granularity and grouping of those rules in an export could be anything based on desired goals.  The goals could be communicated in comments or elsewhere.
- Unit tests in a specific vendor test suite (i.e. `fullstory.ts`) would then validate the behavior of example rules.

@TrevorFSmith any feedback on this approach?


TrevorFSmith:
I wonder if we'd be better serving our customers by writing these rules in Javascript since we're expecting tag manager folks to cut and paste into their pages.

van-fs:
After discussion, I'll move the `ts` rule examples into pure JSON.  This should make it easily consumable from the outside.  A `comment` optional will be added to a `DataLayerRule` to allow descriptions since JSON can be difficult to include comments in.

van-fs:
@TrevorFSmith JFYI, this PR is dependent on #32 and #33.

van-fs:
I've created three sample rules that work with the CEDDL spec and our mock object.  Notably, there are two optionals added to `DataLayerRule`. There's and `id` and `description`.  I thought an `id` would help teams come up with some versioning or naming strategy for their rules.  It also helps us debug since we now have a rule identifier.  `description` may get used less in the field, but it helps us comment our samples since JSON lacks commenting.

van-fs:
@TrevorFSmith ready for review when you are.

PR #32, opened 2020-07-14
Add  suffix operator

van-fs:
This PR adds the `SuffixOperator`, which is needed for suffixing objects with type information before sending to FullStory.  This will place vendor-specific behavior into core, but suffixing is complex enough that it requires a large amount of code to do so and bolting it on from the outside would be a chore.

Since suffixing is required if using FullStory, the `beforeDestination` option will be used to specify that the `suffix` operator run every time rather than requiring specific reference to suffix in the operator chain.

### DIFF
--- a/src/operators/suffix.ts
+++ b/src/operators/suffix.ts
@@ -1,0 +1,172 @@
+import { OperatorValidator, OperatorOptions, Operator } from '../operator';
+
+/**
+ * A SuffixedObject is an object that has had FS types appended to properties.
+ */
+export type SuffixedObject = { [key: string]: SuffixableValue | SuffixedObject };
+
+/**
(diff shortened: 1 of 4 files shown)


src/operators/suffix.ts
```diff
@@ -0,0 +1,172 @@
+import { OperatorValidator, OperatorOptions, Operator } from '../operator';
+
+/**
+ * A SuffixedObject is an object that has had FS types appended to properties.
```
TrevorFSmith, line 4: Since this is a general purpose operator, maybe we can leave out the FS-specific comments?

van-fs: As I re-read it this morning I thought the same.  Technically suffixing extends beyond our needs.

PR #33, opened 2020-07-15
Add insert operator

van-fs:
This PR introduces the `InsertOperator`.  In the prototype, there was a `prepend` operator that added a value to the beginning of a list.  The use case here is adding information to a list of function arguments that are passed to the `destination` function.  So you might add 'Add to Cart' as a literal to FS.event arguments.  Or you might grab the value of `profileID` to be used with FS.identify.  The design was expanded to be more flexible - supporting arbitrary insertion into the list rather than specifically prepending.  The comments in the `InsertOperator` have detail as well.

### DIFF
--- a/src/operators/insert.ts
+++ b/src/operators/insert.ts
@@ -1,0 +1,55 @@
+import { Operator, OperatorOptions, OperatorValidator } from '../operator';
+import { select } from '../selector';
+
+export interface InsertOperatorOptions extends OperatorOptions {
+  select?: string; // a value found using selection syntax
+  value?: boolean | string | number | object; // a given value to insert
+  position?: number; // the location where the value will be inserted
+}
+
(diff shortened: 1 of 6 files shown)


src/operators/insert.ts
```diff
@@ -0,0 +1,55 @@
+import { Operator, OperatorOptions, OperatorValidator } from '../operator';
+import { select } from '../selector';
+
+export interface InsertOperatorOptions extends OperatorOptions {
+  select?: string; // a value found using selection syntax
```
TrevorFSmith, line 5: It's pretty slick that we can mix in selection syntax wherever we need it. Go, us!

van-fs: I love this feature.  It has made the operator chain so much cleaner.

PR #34, opened 2020-07-16
Add QueryOperator

van-fs:
@TrevorFSmith @patrick-fs  While writing the user rules, I had a need to do the following:

```
{
    "id": "fs-identify-ceddl-user-allowed",
    "description": "send only known CEDDL user properties to FS.identify using the profileID as the FullStory uid",
    "source": "digitalData.user.profile[0][(profileInfo.profileID,profileInfo.userName,address.line1,...)]",
    "operators": [
      {
        "name": "flatten"
      },
      {
        "name": "insert",
        "select": "profileID"
      }
    ],
    "destination": "FS.identify"
  }
```

That stretches the ability of the selector syntax.  Do we feel like that is valid syntax?

Alternatively, I was thinking we might have an operator that allows us to perform selection within any step of the operator chain.  It would look something like this:

```
  {
    "id": "fs-identify-ceddl-user-allowed",
    "description": "send only known CEDDL user properties to FS.identify using the profileID as the FullStory uid",
    "source": "digitalData.user.profile[0]",
    "operators": [
      {
        "name": "flatten"
      },
      {
        "name": "query",
        "select": "$[(profileID,userName,line1,line2,city,stateProvince,postalCode,country)]"
      },
      {
        "name": "insert",
        "select": "profileID"
      }
    ],
    "destination": "FS.identify"
  }
```

The flatten step yields an intermediate top level object:

```
{
 profileID: ...,
 userName: ...,
...
}
```

The query operator then picks allowlisted properties from the intermediate object.

There's on noticeable character in the selector, the `$`.  What I found is that I can't use `[(profileID,userName,line1,line2,city,stateProvince,postalCode,country)]` standalone.  I suspect the selector implementation is expecting some property to precede the picking.  So what I've done under the hood is created an intermediate object `{ $: <object>` to allow the picking to succeed. The `$` was part of the original [JSON Path](https://goessner.net/articles/JsonPath/) project so it's not completely foreign to use.  The question is whether it's needed or if we should revise the selector.

If we do go with `QueryOperator`, I want to sync on naming.  In the `InsertOperator` I used `select` as an option.  It's carried over as an option here - as in "running a select query".  That feels natural, and in the future we might have more executable queries so I went with `QueryOperator` versus `SelectOperator`.

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -3,6 +3,7 @@ import {
   FlattenOperator, FlattenOperatorOptions,
   InsertOperator, InsertOperatorOptions,
   SuffixOperator, SuffixOperatorOptions,
+  QueryOperator, QueryOperatorOptions,
 } from './operators';
 import { Operator } from './operator';
 
@@ -10,13 +11,13 @@ import { Operator } from './operator';
  * Declares known, built-in Operators.
  */
 export type BuiltinOperator = typeof FlattenOperator | typeof FunctionOperator | typeof InsertOperator
-  | typeof SuffixOperator;
+  | typeof SuffixOperator | typeof QueryOperator;
 
 /**
  * Declares known, built-in OperatorOptions.
  */
 export type BuiltinOptions = FlattenOperatorOptions | FunctionOperatorOptions | InsertOperatorOptions
-  | SuffixOperatorOptions;
+  | SuffixOperatorOptions | QueryOperatorOptions;
 
 /**
  * OperatorFactory creates instances built-in Operators. Since DataLayerRule can define OperatorOptions at runtime,
@@ -28,6 +29,7 @@ export class OperatorFactory {
     function: FunctionOperator,
     insert: InsertOperator,
     suffix: SuffixOperator,
+    query: QueryOperator,
   };
 
   /**
--- a/src/operators/index.ts
+++ b/src/operators/index.ts
@@ -2,3 +2,4 @@ export * from './flatten';
 export * from './function';
 export * from './insert';
 export * from './suffix';
+export * from './query';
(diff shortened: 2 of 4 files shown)


TrevorFSmith:
Looking at this selector: `profile[0][(profileInfo.profileID,profileInfo.userName,address.line1,...)]`

The "double bracket" syntax of `prop[num][(...)]` isn't currently supported but probably could be. An alternative is to use two rules with the first rule selecting prop[num] and the second rule plucking values.

Using dot notation (or any other selector syntax within a bracket-contained property list) isn't currently supported. I'm hesitant to introduce that sort of recursive parsing and selection because it quickly becomes complex to determine how each individual property's syntax in the list will be resolved. If we can give people multi-step rules to get where they need to go with each step easy to understand then I suspect that more tag managers will pick it up. 

To that point, I like the `query` syntax that @van-fs suggested. Perhaps we could limit it in scope and clarify the name by naming it `find-first` and having it just do a breadth-first search and use the first value that matches the property name? 

Another approach would be to add wildcard or other regex-style syntax to the property list, beyond the currently implemented equality operator of `(prop=value, ...)`. So, we could add something like `(prop~=foo??, ...)` to assign values like `foo23` to `prop`.

van-fs:
Per sidebar commentary, I feel like making the grammar of the selector more complex to support with make it less user friendly.  By delegating some of this additional behavior to an operator, we can minimize the complexity and long-term maintenance of selector.  The operator makes these sophisticated selections more obvious as there is a way to currently debug the rule.  On the table is the usage of `$` and whether it has a role in the selector itself.  I feel that for now `$` scoped to just the QueryOperator is explainable as denoting a "query".

PR #38, opened 2020-07-20
add FullStory CEDDL Cart spec

van-fs:
This PR adds examples for FullStory and CEDDL Cart.  The example JSONs were converted to objects with a `rules` property.  I saw some odd behavior when trying to iterate over the pure list.  Making the examples objects with a `rules` property and a value of an array solved this.  Also added a `getRule` function to make it clearer which rule to test rather than relying on the order in the list.

### DIFF
--- a/test/fullstory.spec.ts
+++ b/test/fullstory.spec.ts
@@ -61,7 +61,7 @@ describe('FullStory example rules unit tests', () => {
     expect(payload.attributes).to.be.undefined;
     expect(payload.job).to.eq('developer'); // verify custom property
 
-    (globalThis as any).digitalData.user.profile[0].job; // remove custom property
+    delete (globalThis as any).digitalData.user.profile[0].job; // remove custom property
   });
 
   it('it should send any CEDDL user property to FS.identify', () => {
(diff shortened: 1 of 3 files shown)


test/fullstory.spec.ts
```diff
@@ -83,4 +93,70 @@ describe('FullStory example rules unit tests', () => {
...
+    const observer = new DataLayerObserver({ rules: [getRule('fs-event-ceddl-cart')], readOnLoad: true });
+    expect(observer).to.not.be.undefined;
+
+    const [eventName, payload] = expectParams(globalMock.FS, 'event');
+    expect(eventName).to.eq('View Cart');
+    expect(payload.cartID).to.eq(cartID);
+    expect(payload.price).to.eq(price);
+    expect(payload.promotion).to.be.undefined;
+
+    (globalThis as any).digitalData.cart.promotion; // remove custom property
```
TrevorFSmith, line 111: Does this need a `delete`?

test/fullstory.spec.ts
```diff
@@ -83,4 +93,70 @@ describe('FullStory example rules unit tests', () => {
...
+
+    const [eventName, payload] = expectParams(globalMock.FS, 'event');
+    expect(eventName).to.eq('View Cart');
+    expect(payload.cartID).to.eq(cartID);
+    expect(payload.price).to.eq(price);
+    expect(payload.attributes).to.eq(attributes);
+    expect(payload.promotion).to.eq('LaborDay2020');
+    expect(payload.items).to.be.undefined;
+
+    (globalThis as any).digitalData.cart.promotion; // remove custom property
```
TrevorFSmith, line 131: `delete`, here too?

PR #40, opened 2020-07-21
Examples rules for converting CEDDL products into FS events.

TrevorFSmith:
Opening this as a draft to get feedback from @van-fs about how to handle arrays like CEDDL's `digitalData.product`.


van-fs:
That's basically what I would have done.  For other data layers I've seen, a PDP page will contain a single product of interest, and the digitalData.products list only has one item.  So either looking at index 0 or -1 yields the same.  I think you could omit the first operator `{"name": "query","select": "$[(category, productInfo)]"}`.  The flatten operator should take those object properties and discard them as part of the flattening process.  One curiosity is what the empty array would be if we sent that to FullStory.  The suffix operator will try to figure out the type and since it's an empty list, it should return an empty string for the suffix type, which means it's dropped from the object before sending to FullStory.

PR #42, opened 2020-07-21
add adobe mocks and rules

van-fs:
This is a draft PR for Adobe mocks and simple eVar based rules.  There's nothing really in core, it's mostly boilerplate and rule writing.  Since Adobe is a lot more generic in its eVars, there's less guidance we can really provide out of the box.  There are some properties like `campaign` that we could grab, but I started with eVars since that was a focus area to explore.


van-fs:
@TrevorFSmith I'm going to convert this to Ready for review.  Not sure how much else I'd change for the current iteration's goals.

PR #45, opened 2020-07-23
add more comparisons to selector

van-fs:
This adds more comparison features to the selector (currently we only support `=`).  Note this uses `eval`.  Alternatively, I can remove the eval and do a switch on the combinations of operators `<=`, `!=`, `>=`, etc.


van-fs:
Eval removed in favor of specific checks on operators.

PR #46, opened 2020-07-24
bug fix for boolean and nulls

van-fs:
I fixed a bug in how the boolean conversion from a `'true'` or `'false'` was done.

The other scenario encountered was if a rule wanted to always try to convert a value to a real or int.  If the value was null, we'd actually leave it as is but console.error that we couldn't convert.  This can get noisy and gives the appearance of an error when none really exists.  So this fix also assumes that a null being converted to an int/real yields a zero value.  It's worth thinking through if there are any undesirable side-effects.  A similar question is whether other type conversion of null (e.g. `string`) should provide a default value such as the empty string.


van-fs:
Also addressed a bug where the splice was leaving the original object and the converted object.  The original should not be included in the returned data.

PR #48, opened 2020-07-28
add FS.event appender

van-fs:
Looks like we had support for the custom log appender but it was not available to declare as an option to `init`.  This PR allows mounting `FS.event` as an appender so that we can pipe logs/errors to FullStory.  Usage would be similar to:

```
window['_dlo_appender'] = {
      log: ({ level: level_int, message: message_str, datalayer: datalayer_str }) => {
        const FS = window[window['_fs_namespace']];
        console.log({ level_int, message_str, datalayer_str })
        if (FS) FS.event('Data Layer Observer', { level_int, message_str, datalayer_str }, 'dataLayerObserver')
      }
    };
```

### DIFF
--- a/src/embed/init.ts
+++ b/src/embed/init.ts
@@ -5,6 +5,10 @@ import { DataLayerObserver } from '../observer';
 /*
 This is where we initialize the DataLayerObserver from this info:
 
+// A custom log appender; a console appender is used if one is not specified
+// Default is null
+window['_dlo_appender'] = null;
+
 // OperatorOptions that is always used just before before the destination
 // Default is null
 window['_dlo_beforeDestination'] = null;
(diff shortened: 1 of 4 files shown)


van-fs:
I'm trying to track down what looks like a race condition in the test.  If I debug and step through it, no issues.  Odd because I did not think we had any async code that would cause this.

van-fs:
@TrevorFSmith Fixed the race condition issue in the test case.  It's weird cause previously it was a closure but it kept saying `FS` wasn't defined in globalThis.  Odd because entry to the test case did not have this issue AND debugging just long enough to mash play fixed it.  The class seems to work better so perhaps some weird scoping issue with mocha??

PR #50, opened 2020-07-30
Add Web Embed Lab Testing

TrevorFSmith:
This PR sets up two new npm run targets: `test:wel-auto-formulate` and `test:wel-runner`

`test:wel-auto-formulate` uses the Web Embed Lab to test the DLO in the context of two example pages, one which is the most basic setup (`examples/basic-embed`) and a new one that tests CEDDL to FullStory rules (`examples/ceddl-fs`).

I'll add docs about setup once I've heard from @van-fs about this general direction, but to run the tests the user needs  to set up two files (both in the `test/web-embed-lab/` dir) that configure the WEL with ngrok and Browserstack keys. Basically, the user has to copy and fill out the *.example files in that dir (removing the `.example` extension).

### DIFF
--- a/test/web-embed-lab/README.md
+++ b/test/web-embed-lab/README.md
@@ -6,7 +6,7 @@ Web Embed Lab (WEL) is an [open source tool](https://github.com/fullstorydev/web
 
 WEL uses Browserstack and ngrok so you'll need to set up a `.env` configuration file or environment variables to give WEL the information it needs.  Read the [environment vars doc](https://github.com/fullstorydev/web-embed-lab/blob/master/docs/ENVIRONMENT_VARS.md) in the WEL repo for details.
 
-You'll also need to copy the `ngrok-auto-formulate-config.yml.example` into `ngrok-auto-formulate-config.yml.example`
+You'll also need to copy the `ngrok-auto-formulate-config.yml.example` into `ngrok-auto-formulate-config.yml.example` and update it with your ngrok API key.
 
 To test the DLO with WEL, there are two npm targets:
 
--- a/test/web-embed-lab/README.md
+++ b/test/web-embed-lab/README.md
@@ -1,0 +1,23 @@
+# Web Embed Lab
+
+The Web Embed Lab (WEL) is an [open source tool](https://github.com/fullstorydev/web-embed-lab) used to test that embedded scripts (like the DLO) don't negatively effect customer sites.
+
+## Configuration
+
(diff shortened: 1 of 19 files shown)


test/web-embed-lab/README.md
```diff
@@ -0,0 +1,23 @@
+# Web Embed Lab
+
+The Web Embed Lab (WEL) is an [open source tool](https://github.com/fullstorydev/web-embed-lab) used to test that embedded scripts (like the DLO) don't negatively effect customer sites.
```
van-fs, line 3: nit Can we drop the "the" when we refer to Web Embed Lab, WEL, and DLO.

test/web-embed-lab/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The WEL uses Browserstack so you'll need to set up a `.env` configuration file or environment variables to give the WEL the information it needs for your Browserstack account.  Read the [environment vars doc](https://github.com/fullstorydev/web-embed-lab/blob/master/docs/ENVIRONMENT_VARS.md) in the WEL repo for details.
+
+To test the DLO with the WEL, there are two npm targets:
+
+## Autoformulate
+
+The following command will capture example sites "freeze dry" them into replayable pages (WEL calls them "page formulas") in which we can run tests.
+
+	npm run test:wel-auto-formulate
```
van-fs, line 15: Let's code fence the commands so they stand out more.

test/web-embed-lab/README.md
van-fs: Is it affect?

van-fs:
Most of it LGTM.  I saw the domain arg was used with ngrok, so I guess that would need to also be configurable or part of the future docs.  Would we also advocate using a real website for any testing/benchmarking purposes?

TrevorFSmith:
> Most of it LGTM. I saw the domain arg was used with ngrok, so I guess that would need to also be configurable or part of the future docs. Would we also advocate using a real website for any testing/benchmarking purposes?

My general plan is to get this PR (which tests that nothing on the tested pages breaks) merged and then in a future PR(s) build out test probes that are specific to rule sets (e.g. CEDDL -> FS). These test probes will allow us to more specifically check that the expected events with expected data arrived.

Regarding testing against 3rd party public sites: What do you think about using separate configs for auto-formulating and running against those sites so that people don't need to bang on them if they just want to test the formulations of the local examples. (again in a future PR)

PR #54, opened 2020-08-03
add function operator doc

van-fs:
Add the function operator docs.

### DIFF
--- a/docs/operator_function.md
+++ b/docs/operator_function.md
@@ -1,0 +1,84 @@
+# Function Operator
+
+The function operator applies a function on a list of arguments and returns the result.  This operator supports the `destination` feature of a data layer rule; however, it can be used within a list of `operators` to perform more sophisticated transformations.
+
+## Options
+
+Options with an asterisk are required.
+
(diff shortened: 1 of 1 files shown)


docs/operator_function.md
```diff
@@ -0,0 +1,84 @@
...
+| `func`* | `string | function` | `undefined` | Function to apply on the operator input list. |
+| `thisArg` | `string | object` | `globalThis` | Object or selector used as the `this` context with the function. |
+
+If a selector is used with `func` or `thisArg`, the selector will be evaluated to find the function or object.
+
+## Usage
+
+## Converting a list of objects to a list of strings
+
+### Rule
```
TrevorFSmith, line 20: Maybe we could start with a dead simple function example and then expand to the CEDDL example? It feels a little overwhelming just because of the size of the input data.

van-fs: @TrevorFSmith What if I just trimmed the input a bit more, take out some of the extraneous properties - enough to demo the point?

PR #63, opened 2020-08-12
Add the rename operator

TrevorFSmith:
The rename operator is pretty basic.

I did have to turn off lint and ts on lines 28 and 30 of rename.ts so a double-check that those are OK would be good.

### DIFF
--- a/src/operators/rename.ts
+++ b/src/operators/rename.ts
@@ -1,0 +1,60 @@
+import { Operator, OperatorOptions, OperatorValidator } from '../operator';
+
+export interface RenameOperatorOptions extends OperatorOptions {
+  properties: { [key: string]: string };
+}
+
+export class RenameOperator implements Operator {
+  static specification = {
+    index: { required: false, type: ['number'] },
+    properties: { required: true, type: ['object'] },
+  };
+
+  readonly index: number;
+
+  readonly properties: object;
+
+  constructor(public options: RenameOperatorOptions) {
+    const { index = 0, properties = {} } = options;
+    this.index = index;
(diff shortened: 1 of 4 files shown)


src/operators/rename.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.properties = properties;
+  }
+
+  handleRename(data: { [key: string]: any }) {
+    const props = Object.getOwnPropertyNames(this.properties);
+    for (let i = 0; i < props.length; i += 1) {
+      const key = props[i];
+      const val = data[key];
+      /* eslint-disable no-param-reassign */
+      delete data[key];
```
van-fs, line 29: Should we use `// eslint-disable-line no-param-reassign` rather than the eslint block comment above?  If the block is used, I believe that disables the rule for subsequent lines in the file.  I think we just make the exclusion specific to this line.

src/operators/rename.ts
```diff
@@ -0,0 +1,60 @@
...
+      // @ts-ignore
```
van-fs, line 30: I noticed the ignore, which I assume is because the `object` type isn't indexable.  Since the validator expects a `string` as the renamed property, should we then assume that the existing property is also a string?  (At least for type declaration in the TS file.)  The `object` type would then give way to `{ [key:string]: string }`.  It looks like any numeric property would get coerced to a string at runtime.  Or was there a non-string use case in mind?

TrevorFSmith: Ok, I've changed the lint specifiers and fixed the typescript complaints. It's ready for another review.

PR #65, opened 2020-08-21
Add shim property change listener

van-fs:
This PR adds the concept of a "monitor" to watch for changes on a property.  I used "monitor" so it wouldn't be confused with implementation, which uses EventListeners.
- The shimming technique applies to all properties in a given object, which would be the result of a selector having been run.  You can optionally not shim `monitor=false` in a rule if you only wanted to `readOnLoad`.
- The specs use NPM deepcopy since I had a few issues with test pollution.
- I had to add `monitor: false` to all our previous observer.spec rules since the default behavior was true and that had side-effects.  That and a lint:fix caused more changes than normal.
- I added some selector.spec tests because I'm explicitly relying on dot notation select to return a reference to an object and not a copy.

### DIFF
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -5,7 +5,7 @@ import { Logger, LogAppender } from './utils/logger';
 import { FunctionOperator } from './operators';
 import Monitor from './monitor';
 import ShimMonitor from './monitor-shim';
-import { select } from './selector';
+import { select, parsePath, ElementKind } from './selector';
 import { DataLayerEventType } from './event';
 
 /**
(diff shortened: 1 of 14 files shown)


src/observer.ts
```diff
@@ -181,10 +197,38 @@ export class DataLayerObserver {
...
+    const monitors: Monitor[] = [];
+
     try {
       const handler = this.addHandler(source);
       handler.debug = !!debug;
 
+      try {
+        if (monitor) {
+          // use select to get the target with the desired properties
+          // NOTE using [()] to pick, etc returns a copy and not the actual data layer reference
```
TrevorFSmith, line 209: I feel like this is going to create a bit of mystery. I wonder how we could surface this during rule writing so that tag managers understand what's happening.

src/observer.ts
```diff
@@ -181,10 +197,38 @@ export class DataLayerObserver {
...
+          const target = select(source);
+
+          if (target) {
+            const braketPos = source.lastIndexOf('[(');
```
TrevorFSmith, line 213: I get nervous when I see string parsing of selectors happening outside of selector.ts. Perhaps a better way would be to export the `parsePath` method from that file? It returns a (memoized) Path instance and we could add the string parsing and formatting logic to that class.

src/observer.ts
```diff
@@ -181,10 +197,38 @@ export class DataLayerObserver {
...
+
+            // find the path to the actual reference in the data layer by selecting with a path
+            const path = source.substring(0, braketPos === -1 ? source.length : braketPos);
+            const ref = select(path);
+
+            Object.getOwnPropertyNames(target).forEach((property) => this.addMonitor(ref, property, source));
+
+            window.addEventListener(DataLayerEventType.PROPERTY, (e: Event) => handler.handleEvent(e as CustomEvent));
```
TrevorFSmith, line 221: It would be handy to be able to clean up these global listeners at the end of a test to prevent mysterious event bleeding. Alternatively, we could somehow tag events with a per-Observer unique id so observers can filter out events meant for others.

src/observer.ts
van-fs: @TrevorFSmith I exported the `parsePath` and `ElementKind` so I could rely on it to walk the selector path better.  The one aspect that is not internal to `selector.ts` is the `Brackets` expose the `Op` `raw` value, which is something like `(cartID,price)` but not the surrounding bracket chars.  I could enhance `Bracket` to have this additional "raw brackets" value or we can keep `ref = select(source.substring(0, source.indexOf(`[${brackets.op.raw}]`)));`.  If we go with "raw brackets" as a new instance member in `Brackets`, what might we call that?

van-fs:
@TrevorFSmith  the revisions are done.

There's a lot that conceptually overlaps with the previous PR version.  Some notable changes:

- `utils/mocha.ts` has an `ExpectObserver` that should make it easier to deregister the window listeners, which are a side effect of the creating an observer.  There's also an `expectEventListener` that similarly checks and removes window event listeners.

- `selector.ts` had two exports added to expose some of the parsing so I'm not doing duplicating.

- There's a new class called `DataLayerTarget`.  This internalizes some of the semantics for source, target, path, selector, object.  Unfortunately, over time, these terms tend to get used interchangeably and led to my own confusion over what a variable implied.  So no we have:

  * `source` a selector defined in a rule that describes and object in the data layer and desired data to be obtained
  *  `path` is a subset of a selector and uses dot or index notation to traverse an object; it should return a reference to an object or function in the data layer
  * `object` the actual object in the data layer found via a `path`
  * `target` the "thing" in the data layer we want to observe; `DataLayerTarget` bundles all of these up so it's clear the above have distinct purposes but are related

- `Monitor` and `ShimListener` are more or less the same, but I tried to align naming to suit the above.  Shims now get added to everything, and let's keep an eye on the performance impact.  Right now, it's not shimming multiple levels of an object, but that's feels like an expectation someone will have.

- `window.dispatchEvent` no longer emits a generic `_dlo_DataLayerProperty` type.  This is now very tightly aligned to the `path` concept, and emits a type like `datalayerobserver/<some_real_path>`.  So every `DataHandler` is watching a `path` and it's up to the handler to decide what to do.  If the handler runs a source select and doesn't get any data, it doesn't do anything.  This should prevent FS.events with eventNames but empty payloads. The upside here is that a `DataHandler` now internalizes it's `window.addEventListener` registration (we ruminated on this in Slack).

PR #66, opened 2020-08-27
NPM module packaging

van-fs:
Basic mechanics of packaging for NPM.  I'd like to get the functional shimming in so that I can test more in the App Insights demo.  Once I can validate the usage, we can do a PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,15 +3,18 @@
   "version": "1.0.0",
   "description": "Monitor, transform, and send data layer content to FullStory",
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
     "build": "rollup --config",
+    "build:module": "tsc",
     "watch": "rollup --config -w",
     "examples": "npm run build && http-server",
     "test": "nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts test/**/*.spec.ts",
     "test:wel-auto-formulate": "npm run examples & cd ./test/web-embed-lab && ./auto-formulate.sh",
     "test:wel-runner": "npm run build && cd ./test/web-embed-lab && ./run-tests.sh",
     "lint": "eslint --ext .ts src/** test/mocks/** test/utils/** test/*.ts",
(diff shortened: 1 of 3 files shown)


TrevorFSmith:
Ok, shims are merged.

van-fs:
Tests are passing with the refactor for programmatic observation.  Couple of points:
- I had to mod the tsconfig to remove `outDir` and use that as a CLI flag for the `build:module` script.  This is an active [rollup bug](https://github.com/rollup/plugins/issues/287) and having the `outDir` in tsconfig caused rollup to fail.
- I went back and enhanced `DataLayerTarget` to handle the variety of conditions we need to support.  I expect we'll run into something eventually, but this is a solid starting point.  I created a spec for DataLayerTarget to be more clear about the expectations.
- I simplified the API points in `DataLayerObserver`.  There's `registerRule` for our typical window-based data layers.  This uses a new function `registerTarget` that works with custom apps.  One of the goals was to move some of the basic mechanics of set up out of the `processRule` function, which had a lot of try catch behavior that was a code smell.

Programmatic usage looks a bit like this, which we'll need to document at the end of the sprint.
```javascript
import { ApplicationInsights } from '@microsoft/applicationinsights-web';
import { DataLayerObserver, DataLayerTarget } from '@fullstory/data-layer-observer';
import * as FullStory from '@fullstory/browser';

...

    // initialize App Insights
    this.appInsights = new ApplicationInsights({
      config: { instrumentationKey: 'a08f3f2d-9884-4437-b6ec-c835d3d58d82' }
    });
    this.appInsights.loadAppInsights();

    // initialize Data Layer Observer (set previewMode to false in production)
    this.observer = new DataLayerObserver({ beforeDestination: { name: 'suffix' }, previewMode: false });

    try {
      // target the App Insights trackEvent function
      const target = new DataLayerTarget(this.appInsights, 'trackEvent', 'appInsights');

      // connect trackEvent to FS.event
      this.observer.registerTarget(target, [], FullStory.event);
    } catch (err) {
      console.error(err);
    }
```

TrevorFSmith (approved):
LGTM

PR #68, opened 2020-08-27
Add function shim

van-fs:
Initial function shim implementation added.  I'm going to revisit and cleanup.  The `DataLayerTarget` and `DataLayerObserver.addMonitor` are a bit messy.  You have to walk back up to the parent to get a reference to it and the `property` (i.e. function).

### DIFF
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -116,7 +116,7 @@ export class DataLayerObserver {
    * Adds monitor to a target in the data layer. If a monitor already exists, calling this
    * function will result in a no-op.
    * @param target to add monitors into
-   * @param property to monitor; if not property is given, the monitor is added to the target itself
+   * @param property to monitor; if property is not given, the monitor is added to the target itself
    */
   static addMonitor(target: DataLayerTarget) {
     const {
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -6,6 +6,7 @@ import { FunctionOperator } from './operators';
 import Monitor from './monitor';
 import ShimMonitor from './monitor-shim';
 import DataLayerTarget from './target';
+import { select } from './selector';
 
 /**
  * DataLayerConfig provides global settings for a DataLayerObserver.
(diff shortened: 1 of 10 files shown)


van-fs:
Added functional shimming.  I created a `MonitorFactory` to internalize some of the creation/removal of monitors and logic of when it's valid to create.  The related logic was moved out of `DataLayerObserver`.  I also updated `DataLayerTarget` with additional metadata about the "subject" because the way we shim functions requires parent information that was not present previously.

src/observer.ts
```diff
@@ -119,18 +116,22 @@ export class DataLayerObserver {
    * Adds monitor to a target in the data layer. If a monitor already exists, calling this
    * function will result in a no-op.
    * @param target to add monitors into
-   * @param property to monitor
+   * @param property to monitor; if not property is given, the monitor is added to the target itself
```
TrevorFSmith, line 119: nit: s/if not property is given/if property is not given/

test/shim-monitor.spec.ts
```diff
@@ -63,6 +66,49 @@ describe('ShimMonitor unit tests', () => {
     globalMock.digitalData.cart.cartID = 'cart-5678';
   });
 
+  it('it should emit args from function calls', (done) => {
```
TrevorFSmith, line 69: It's probably also worth testing that functions which throw exceptions are nicely handled.

van-fs: I went back and looked at the `it should emit args but trap function exceptions` test case:

- The `done` in the event listener tests whether the actual handler was invoked, which should always be the case since `emit` is called prior to the function's `apply`.  `window.dispatchEvent` executes synchronously so `done()` gets called and the next bullet point's code runs.
- The `expect(ret).to.eq(null);` executes next, and this checks sequential lines of code are not impeded by an exception.

The `emit` body in Monitor is wrapped in a try catch block.  I went back and wrote another test `it should emit args and not fail because of dispatch related exceptions` which ensures that if the handler or DLO code were to throw an exception that they are caught and the function call is not impeded.  This is verified by `expect(ret).to.eq(true);` and the fact that the test exits.

PR #69, opened 2020-08-28
Add the fan out operator.

TrevorFSmith:
The fan out operator allows us to specify a set of properties on a source that each should have the remaining rules applied to them. It needs a special case in the Handler so it's a bit different than the other operators.

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -6,20 +6,23 @@ import {
   ConvertOperator, ConvertOperatorOptions,
   QueryOperator, QueryOperatorOptions,
   RenameOperator, RenameOperatorOptions,
+  FanOutOperator, FanOutOperatorOptions,
 } from './operators';
 import { Operator } from './operator';
 
 /**
  * Declares known, built-in Operators.
  */
-export type BuiltinOperator = typeof ConvertOperator | typeof FlattenOperator | typeof FunctionOperator |
-  typeof InsertOperator | typeof SuffixOperator | typeof QueryOperator | typeof RenameOperator;
+export type BuiltinOperator = typeof ConvertOperator | typeof FlattenOperator
+  | typeof FunctionOperator | typeof InsertOperator | typeof SuffixOperator
+  | typeof QueryOperator | typeof RenameOperator | typeof FanOutOperator;
 
 /**
  * Declares known, built-in OperatorOptions.
  */
-export type BuiltinOptions = ConvertOperatorOptions | FlattenOperatorOptions | FunctionOperatorOptions
-  | InsertOperatorOptions | SuffixOperatorOptions | QueryOperatorOptions | RenameOperatorOptions;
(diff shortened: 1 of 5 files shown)


src/factory.ts
```diff
@@ -34,6 +37,7 @@ export class OperatorFactory {
     suffix: SuffixOperator,
     query: QueryOperator,
     rename: RenameOperator,
+    'fan-out': FanOutOperator,
```
van-fs, line 40: nit, but should we just use `fanout` without any special chars?

src/handler.ts
```diff
@@ -127,6 +140,7 @@ export default class DataHandler {
       } catch (err) {
         Logger.getInstance().error(`Operator ${name} failed for ${path} at step ${i}`,
           path);
+        // @eslint-disable-next-line
         console.error(err.message);
```
van-fs, line 144: Did the console log sneak in or do we need it?  If the latter,  should we `Logger.getInstance().error`?

src/operators/fan-out.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  readonly index: number;
+
+  readonly properties: string[];
+
+  constructor(public options: FanOutOperatorOptions) {
+    const { index = 0, properties = [] } = options;
+    this.index = index;
+    if (typeof properties === 'string') {
+      this.properties = [properties];
```
van-fs, line 21: I think this should `properties.split(',')` and assign, right?

test/operator-fan-out.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  it('it should fan out properties', () => {
+    const observer = new DataLayerObserver({
+      rules: [
+        {
+          source: 'item',
+          operators: [
+            { name: 'fan-out', properties: ['left-hand', 'right-hand', 'list'] },
+            { name: 'flatten' },
+          ],
+          destination: 'globalThis.handleDestination',
```
van-fs, line 42: `destination` will actually accept a function too.  In the observer spec, I have:

```
let changes = [];
...
destination: (...data: any[]) => {
            // NOTE use a local destination to prevent cross-test pollution
            changes = data;
          },
```

Nothing functional incorrect, just in case you wanted to keep the destination and result localized.

src/factory.ts
TrevorFSmith: Everywhere else in the code I treat fan out as two words. Do you think it's problematic to use skewer-case?

src/handler.ts
TrevorFSmith: I'm not sure when that error arrived in the code but it doesn't seem necessary to me. I'll delete.

src/factory.ts
van-fs: Not problematic; keep it.

PR #71, opened 2020-09-01
Allow the function operator to return null to abort a rule handler.

TrevorFSmith:


TrevorFSmith:
> LGTM. So a function that returns undefined or null essentially has no data to pass along; hence, it makes sense to stop. Correct? Is there any benefit of `typeof val === 'undefined'` versus a `val === undefined`?

I started using `typeof val === 'undefined'` in general because unlike `val === undefined` it doesn't throw an exception if `val` has not been declared. In this situation we declare `val` so it's equivalent to `val === undefined`.

PR #72, opened 2020-09-04
Rules to send Google tags to Fullstory

TrevorFSmith:
In addition to the rules, tests, and example page for google tags, this PR has a couple of tweaks to how we handle rules that are both read-on-load and monitor of array sources.
The monitor also checks that a property is configurable before trying to monitor it.


van-fs (approved):
LGTM; only nit is that Google typically uses `dataLayer` as their convention, but we can change that in a follow on review.

PR #75, opened 2020-09-17
WEL rule probe and separate test "site" 

TrevorFSmith:
I created a [WEL test probe](https://github.com/fullstorydev/web-embed-lab/blob/master/docs/TEST_PROBE_API.md) and a testing page that work together to ensure that the rules ran, that the right number of results were created, and that the results have expected parameters.

In the future we can flesh this out to test the intricacies of each operator but this satisfies the task which is to make sure that the DLO successfully evaluates rules.

### DIFF
--- a/test/web-embed-lab/test-probes/dlo-rules/test.js
+++ b/test/web-embed-lab/test-probes/dlo-rules/test.js
@@ -24,7 +24,7 @@ class DLORulesProbe {
     if(!basis) return { passed: true }
     const results = {
       passed: true,
-      results: window._testRuleResults || null, // List of rule results
+      results: window._testRuleResults, // List of rule results
       failures: []
     }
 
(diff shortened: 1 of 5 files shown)


test/web-embed-lab/test-probes/dlo-rules/test.js
```diff
@@ -0,0 +1,59 @@
+/**
+DLORulesProbe tests that a well known array contains expected rule results
+*/
+class DLORulesProbe {
+  /**
+  @return {Object} data collected when the target embed script *is not* loaded
+  @return {Object.success} always true
+  */
+  async gatherBaselineData(){
+    console.log('DLO rules baseline')
```
van-fs, line 10: nit: semicolon missing.  I suspect our linter isn't working on these tests.

test/web-embed-lab/test-probes/dlo-rules/test.js
```diff
@@ -0,0 +1,59 @@
...
+          results.failures.push('The parameters array must contain arrays like [index, value]')
+        }
+      }
+    }
+
+    return results
+  }
+}
+
+window.__welProbes['dlo-rules'] = new DLORulesProbe()
```
van-fs, line 59: same nit; semicolon.

van-fs (approved):
LGTM; two super-nits that have no functional impact.

test/web-embed-lab/test-probes/dlo-rules/test.js
TrevorFSmith: Yep, the WEL isn't using the linter (or semicolons) so the new test is consistent with the other test probes from the WEL.

PR #76, opened 2020-09-22
Rule performance

TrevorFSmith:
I created a [quick performance doc](https://github.com/fullstorydev/fullstory-data-layer-observer/blob/tfs/wel-perf/docs/performance.md) that explains that there are a couple of load test pages (one basic and one more complex) that I used to gather performance snapshots in Chrome and Firefox.

TL;DR each rule takes ~0.07ms to initialize and run.

This isn't hooked up to the WEL tests, mostly because the timing is so fast that it would be lost in the relatively slow noise of the debugger snapshots used by the WEL test probes.


TrevorFSmith:
Ok, @van-fs I pulled in your copy and new image from the gdoc into the performance doc markdown. So, this is ready for review.

van-fs (approved):
LGTM

PR #77, opened 2020-09-24
Magical unshift

TrevorFSmith:
Make the magical monitoring of arrays include `unshift` in addition to `push`.

This is the first time that a single input rule can result in multiple running rules. Instead of converting a rule that watches an array into a rule that watches that array's `push` we now do that and *in addition* add a rule to watch `unshift`.

### DIFF
--- a/examples/basic-embed/index.html
+++ b/examples/basic-embed/index.html
@@ -11,7 +11,8 @@
           pageInfo: {
             pageID: '1234'
           }
-        }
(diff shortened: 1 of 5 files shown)


TrevorFSmith:
@van-fs Ok, this refactor based on our conversation feels like an improvement.

van-fs (approved):
LGTM

PR #79, opened 2020-09-28
Add Google Tag eCommerce rules

TrevorFSmith:
Added rules (with example page and tests) that hand the [Google Tag eCommerce](https://developers.google.com/tag-manager/enhanced-ecommerce) data layer objects to Fullstory.


van-fs (approved):
LGTM

PR #82, opened 2020-10-05
Update Google Enhanced Ecommerce rules

van-fs:
@TrevorFSmith Adding a draft PR to deal with the fan-out issue with Google EC rules.  I created an uber rule to look at the `action` property.  From what I can tell, every Google EC invocation should be adding an `action` property to `actionField`, which should be a result of Google `ga('ec:setAction', 'detail');` calls.

```
    {
      "id": "fs-ga-e-commerce-action",
      "source": "dataLayer",
      "operators": [
        { "name": "query", "select": "$.ecommerce" },
        { "name": "flatten" },
        { "name": "query", "select": "$[?(action)]" },
        { "name": "query", "select": "$[!(products,impressions,promotions)]" },
        { "name": "convert", "properties": "revenue,tax,shipping", "type": "real" },
        { "name": "convert", "properties": "step", "type": "int" },
        { "name": "insert", "select": "action" }
      ],
      "destination": "FS.event",
      "readOnLoad": true,
      "monitor": true
    },
```

So this will flatten `ecommerce` and then grab the `action` field.  Unfortunately, we can't easily go straight to `action` via a pick because the path is `ecommerce.<the action>.actionField`.  As a result, the flatten throws out `<the action>` property.

The previous rules you had can then operate on the products/promotions using fan-out.  I've used the naming convention to match the [Google action](https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#action-types).  A `click` on a product yields the event name `product_click` and similarly a promotion is `promotion_click`.

Also, I changed `View Page Type` to `pageview`.  I've pulled back a bit on examples for impressions as they can get quite noisy from what I'm seeing.  One site has 45 items in the array.  So maybe we revisit impressions or possibly look into a `limit` feature for fan-out.

I'm going to do a bit of integration testing but have a look at where things are headed.


van-fs:
Looking at some sites, the view cart activity is going to emit all products in the cart.  With a rate limit of 10 events per second in FullStory, we're likely to cause problems even for this use case.  I think from an educational standpoint the rules make sense but from a production deployment, it may be a footgun.  If the `actionData` object is being correctly used, the aggregate data is more important and emitted far less.

TrevorFSmith:
Looking at [these docs](https://developers.google.com/tag-manager/enhanced-ecommerce), I don't see `action` but I do see `actionField`:
[Product clicks](https://developers.google.com/tag-manager/enhanced-ecommerce#product-clicks) has `actionField` but [Adding a product to the cart](https://developers.google.com/tag-manager/enhanced-ecommerce#add) does not.

van-fs:
I skewed a bit back to the previous samples.  There are two rules: one that checks for `actionField` and the other that checks for `products`.  Each rule pair is specific to the `action` value seen in the path - like `ecommerce.detail.actionField`.  Each action begets the name of the custom event.  The `detail` action is a `detail` event.  The product association is `detail_product` so it should list nicely below the related action in the custom event dropdown.

TrevorFSmith (approved):

PR #86, opened 2020-10-09
Add CEDDL Event Rule and Insert defaultValue

van-fs:
Looks like I never had an `event` rule - probably because we didn't have fan-out at the time.  Adding a rule for it.  

The spec says that the `eventName` property must be present; however, I've already found a data layer that is not following that guidance.  To maximize using the real event's name with non-compliant data layers, I added a `defaultValue` fallback for the `insert` operator.

### DIFF
--- a/src/operators/insert.ts
+++ b/src/operators/insert.ts
@@ -2,6 +2,7 @@ import { Operator, OperatorOptions, OperatorValidator } from '../operator';
 import { select } from '../selector';
 
 export interface InsertOperatorOptions extends OperatorOptions {
+  defaultValue?: boolean | string | number | object; // a default value if the select fails
   select?: string; // a value found using selection syntax
   value?: boolean | string | number | object; // a given value to insert
   position?: number; // the location where the value will be inserted
@@ -26,6 +27,7 @@ export class InsertOperator implements Operator {
   }
 
   static specification = {
+    defaultValue: { required: false, type: ['boolean,string,number,object'] },
     index: { required: false, type: ['number'] },
     select: { required: false, type: ['string'] },
     value: { required: false, type: ['boolean,string,number,object'] },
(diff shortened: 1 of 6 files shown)


src/operators/insert.ts
```diff
@@ -26,20 +27,30 @@ export class InsertOperator implements Operator {
...
 
     if (selection && value !== undefined) {
-      throw new Error('function operator has both \'select\' and \'value\' options set');
+      throw new Error('insert operator has both \'select\' and \'value\' options set');
     }
 
-    const insertedValue = value || select(selection!, data[this.index]);
+    let insertedValue = value || select(selection!, data[this.index]);
+
+    if (insertedValue === undefined && defaultValue) {
```
TrevorFSmith, line 46: If `defaultValue` is a boolean false then wouldn't this fail to assign it? Perhaps this needs to check whether it's undefined?

test/operator-insert.spec.ts
```diff
@@ -67,6 +67,22 @@ describe('insert operator unit tests', () => {
     expect(obj).to.eq(data[0]);
   });
 
+  it('it should insert using defaultValue if selection syntax fails', () => {
+    const data: any[] = [{ foo: 'foo', bar: [{ baz: 'baz' }] }];
+    const operator = new InsertOperator({ name: 'insert', select: 'bar[0].bazzz', defaultValue: 'default' });
```
TrevorFSmith, line 72: Maybe add a few tests for `defaultValue` of each accepted type (string, boolean, object, etc)?

TrevorFSmith (approved):
Two comments about the same thing, which is that I suspect passing in boolean `false` for `defaultValue` won't work as expected. Other than that it LGTM.

src/operators/insert.ts
van-fs: Yes, good catch.

test/operator-insert.spec.ts
van-fs: Done

PR #95, opened 2020-10-19
Error message cleanup

### LINKED ISSUE #56: Bubble up error messages
It appears the higher layers of DLO are masking lower level errors.  For example, an incorrect syntax in a selector yields:

```
Failed to create data handler for rule  (digitalData.user.profile[0)
```

Ideally the selector parsing failure should be reflected in the observer logic.  The same could occur if a validation of a rule's operator were to fail.

van-fs:
This closes #56 and cleans up/bubbles up a lot of the error messages.  I took the approach:
- If DLO is calling a `log` event, a shorter `LogMessageType` was used (e.g. `Empty Event`).  It's a mix of trying to re-use error messages and mimic what's done in [FS API error kinds](https://help.fullstory.com/hc/en-us/articles/360020624394#API%20Error%20Kinds).
- The above lacks a bit of clarity, and when an `Error` is constructed, you can use `Logger.format` and `LogMessage` to create a more descriptive message.  Anything in `err.message` gets put into the context object as `reason_str: <message>`.  This was used heavily in `selector.ts` to be more granular with parsing related errors.

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -1,4 +1,4 @@
-import { Logger } from './utils/logger';
+import { Logger, LogMessageType } from './utils/logger';
 import { Operator } from './operator';
 import { DataLayerDetail, createEventType } from './event';
 import DataLayerTarget from './target';
@@ -53,7 +53,7 @@ export default class DataHandler {
     const { path } = this.target;
 
     if (value === undefined && args === undefined) {
-      Logger.getInstance().warn(`${path} emitted no data`, path);
+      Logger.getInstance().warn(LogMessageType.EventEmpty, { path });
     } else if (type === createEventType(path)) {
       if (value) {
         // debounce events so multiple, related property assignments don't create multiple events
@@ -75,7 +75,7 @@ export default class DataHandler {
         this.handleData(args || []);
       }
     } else {
-      Logger.getInstance().warn(`EventListener received unexpected event ${type}`, path);
+      Logger.getInstance().warn(LogMessageType.EventUnexpected, { path });
     }
   }
 
@@ -129,8 +129,7 @@ export default class DataHandler {
 
         this.runDebugger(`[${i}] ${name} output ${stats}`, handledData, '  ');
       } catch (err) {
-        Logger.getInstance().error(`Operator ${name} failed for ${path} at step ${i}`,
-          path);
+        Logger.getInstance().error(LogMessageType.OperatorError, { operator: name, path, reason: err.message });
         return null;
       }
     }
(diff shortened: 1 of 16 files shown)


src/handler.ts
```diff
@@ -53,7 +53,7 @@ export default class DataHandler {
     const { path } = this.target;
 
     if (value === undefined && args === undefined) {
-      Logger.getInstance().warn(`${path} emitted no data`, path);
+      Logger.getInstance().warn(LogMessageType.EventEmpty, { path });
```
patrick-fs, line 56: I wonder if it makes sense to have two different functions: `warn` which takes a raw string and `warnf` which takes a tokenized string and a variable list of parameters.

van-fs: I like the idea of a formatted `warn` or `error`: `warnf` and `errorf` respectively.  I started this PR with the question, "Where's the sprintf in JS?"  Line 56 above is actually the behavior of a message with a context.

```javascript
interface LogContext {
  rule?: string;
  source?: string;
  operator?: string;
  path?: string;
  property?: string;
  selector?: string;
  reason?: string;
}
```

The spirit of `LogContext` is to provide well-defined, critical information related to debugging.  Depending on scope, some of these members are unavailable, but at all times, at least one of them is likely to be available.  As such, the developer should populate as much of `LogContext` as possible.  `LogContext` also provided a way to fix inconsistencies in string messages and the fact that these error messages become longer (less manageable) the more context we add.

This PR added a `Logger.format` to re-use error messages with slight tweaks for clarity.  For example, `You forgot the )` and `You forgot the (` could be a template with substitution.  The `Logger.format` does not re-use the context though.  I was thinking that `format`'s job is to provide very specific information.  The `)` or `(` variable portion of `You forgot the` would need a member defined in context.  `token` could be added, but then I was concerned that `LogContext` fills up with other stuff and loses that narrow scope.

One solution could be we go with an `errorf(message: string, context: LogContext, ...substitutions?: string[])`.  One issue is that some of the code like `selector.ts` is so deep that it is very close to being completely unaware of context.  If we remove the context param, then the `errorf` feels a bit like `format`.

Open to alternatives or perspective from @TrevorFSmith.

van-fs:
I added another commit to remove the vestigial use of `string | LogContext`.  The `string` option was there from when all DLO used was the `path` as a primary identifier.  With the variety of information now, it seems to make sense to simply remove that as an option altogether and promote the usage of `LogContext` exclusively.  The latter is still optional though I wonder if we want to require `LogContext` for any error or warnings to promote including this info.

TrevorFSmith (approved):

PR #96, opened 2020-10-23
Fix build error

van-fs:
A TS compile-time error was introduced in my last PR to clean up error messages.  Unfortunately, it looks like we don't tsc in pre-commit.  This PR address the actual bug.  @patrick-fs, thoughts on simply adding `tsc` to the `pre-commit` hook?  Would the solution be that easy?

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "npm run build && lint-staged"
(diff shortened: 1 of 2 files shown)


van-fs:
It appears so, I'm going to add that to guard against this in the future.

van-fs:
Also added `npm run build` to the `pre-commit` hook.

patrick-fs (approved):
LGTM

PR #99, opened 2020-10-25
Add google analytics all purpose rule support

van-fs:
This PR adds a bit more all-purpose support to Google Analytics.  While we have Enhanced Ecommerce rules, the basic Google Analytics usage in `ga.js`, `analytics.js`, and `gtg.js` needs to also be supported.  To that end, the rules were updated and a new comparison syntax to support "does not start with" was added.  This allows us to ignore noisy GA events like `gtm.dom` and `gtm.js`.

Google's newer `gtg.js` actually behaves a little differently.  Rather than inserting an object into the data layer, it adds `Arguments`.  This required a bit of creativity to leverage our existing operators.  It's worth documenting as an advanced operator tutorial, but I'll get to that at a later date.  I want to PR the support so that we can add GA as a souq ruleset.

### DIFF
--- a/examples/rules/google-tags-fullstory.json
+++ b/examples/rules/google-tags-fullstory.json
@@ -253,7 +253,7 @@
       "source": "dataLayer",
       "operators": [
         { "name": "query", "select": "$[?(event!^gtm)]" },
-        { "name": "query", "select": "$[!(gtm.uniqueEventId)]" },
+        { "name": "query", "select": "$[!(ecommerce,gtm.uniqueEventId)]" },
         { "name": "insert", "select": "event" }
       ],
       "destination": "FS.event",
--- a/examples/rules/google-tags-fullstory.json
+++ b/examples/rules/google-tags-fullstory.json
@@ -12,6 +12,21 @@
       "monitor": true
     },
     {
+      "id": "fs-gtg-event",
+      "source": "dataLayer",
+      "operators": [
+        { "name": "query", "select": "$[(0,1,2)]" },
+        { "name": "query", "select": "$[?(0=event)]" },
+        { "name": "flatten" },
+        { "name": "query", "select": "$[?(0=event)]" },
+        { "name": "rename", "properties": { "0": "gtgCommand", "1": "gtgAction" } },
+        { "name": "insert", "select": "gtgAction" }
+      ],
+      "destination": "FS.event",
+      "readOnLoad": true,
+      "monitor": true
+    },
+    {
       "id": "fs-ga-e-commerce-detail-action",
       "source": "dataLayer",
       "operators": [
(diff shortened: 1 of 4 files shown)


examples/rules/google-tags-fullstory.json
```diff
@@ -11,6 +11,21 @@
...
       "monitor": true
     },
+    {
+      "id": "fs-gtg-event",
+      "source": "dataLayer",
+      "operators": [
+        { "name": "query", "select": "$[(0,1,2)]" },
+        { "name": "query", "select": "$[?(0=event)]" },
+        { "name": "flatten" },
+        { "name": "query", "select": "$[?(0=event)]" },
```
TrevorFSmith, line 21: Is this `query` using an array index?!

src/selector.ts
```diff
@@ -56,7 +56,7 @@ class OpProp {
     for (let i = 0; i < raw.length; i += 1) {
       const codePoint = raw.charCodeAt(i);
       // the codePoint appears to be some form of comparison operator we support
-      if (codePoint === 33 || (codePoint >= 60 && codePoint <= 62)) {
+      if (codePoint === 33 || (codePoint >= 60 && codePoint <= 62) || codePoint === 94) {
```
TrevorFSmith, line 59: Nit: It would be nice to put inline comments for which code point is which character.

test/selector.spec.ts
```diff
@@ -89,6 +91,7 @@ describe('test selection paths', () => {
     expect(() => { new Path('films[?(dot.prop=1.0)]'); }).to.not.throw();
     expect(() => { new Path('films[?()]'); }).to.throw(Error);
     expect(() => { new Path('films[?]'); }).to.throw(Error);
+    expect(() => { new Path('myEvent[?(event!^gtm.)]'); }).to.not.throw();
```
TrevorFSmith, line 94: We should test that `=^` and `!^` match property names as expected.

TrevorFSmith (approved):
We need a couple of tests that the new property selection operators work as expected. Other than that (and one documentation nit) this LGTM.

test/selector.spec.ts
```diff
@@ -220,6 +223,12 @@ describe('test selection paths', () => {
     expect(select('favorites[?(number!=25)]', testData)).to.be.undefined;
     expect(select('favorites[?(number!=12)]', testData)).to.eq(testData.favorites);
     expect(select('favorites[?(number+=25)]', testData)).to.be.undefined;
+    expect(select('gtmEvent[?(event!^gtm.)]', testData)).to.be.undefined;
```
van-fs, line 226: @TrevorFSmith Here are the individual tests for new `!^` and `=^` comparisons.

TrevorFSmith (approved):

examples/rules/google-tags-fullstory.json
van-fs: This where it's interesting, the `Arguments` object is essentially an array.  What we do is pick properties `0`, `1`, `2`, which has the effect of getting the values at array index 0-2.  That returns a POJO with 

```
{
 "0": <the 0 element>,
 "1": <the  1 element>,
 "2": <the 2 element>,
}
```

It then queries the "0" property, which is set to the string `event`.

PR #105, opened 2020-11-12
Expose logging level config option

van-fs:
We've always had the ability to set and leverage `LogLevel` to control debug messages.  This PR exposes that ability as a config option.  External configurability allows souq to turn off logging if desired.


TrevorFSmith (approved):
LGTM

PR #109, opened 2021-01-06
Ability to compare undefined

van-fs:
The Google Enhanced Ecommerce ruleset and Event Measurement rulesets are very similar.  When used together, I think there a definite possibility we might record an event for each if an Enhanced Ecommerce hit fires.  The reason is that EM is essentially a superset of EC.  To better ignore EC events when also using the EM ruleset, I added a check that allows us to see if a property is/is not undefined.  This way we can check if `ecommerce` is defined in the event and if so, ignore it assuming the EC ruleset is also deployed.


TrevorFSmith (approved):
LGTM

PR #110, opened 2021-01-06
String not equal bug

van-fs:
While doing #109 I happened to notice there was no `!=` for strings in `select`.  I also saw line 198 in the test looked like a copy/paste.  Wondering if `!=` has gone unnoticed as a bug.  @TrevorFSmith , does the previous line 399 jog any memory of why it doesn't conform to all the other if blocks.  This is the only one that used a `!=` comparison.  Feels out of place.


TrevorFSmith (approved):
LGTM

TrevorFSmith:
Yep, I think it's a copy-pasta bug.

PR #113, opened 2021-01-19
Convert when spaces are provided

van-fs:
Quick fix for an issue found with `convert` operator and spaces in the property list.


patrick-fs (approved):
LGTM

PR #116, opened 2021-02-15
Add enumerate feature to convert operator

van-fs:
This PR addresses a few issues seen with Tealium users.
- The sample rule needs frequent updates to convert strings to numbers
- `order_` information was not included in the original sample
- The `convert` operator could do a better job of intelligently converting to numbers

An `enumerate` option was added to `convert`.  The intent is that if specified, `convert` will try to automatically convert strings to numbers.  There was a fair amount of cleanup needed in convert, and I broke the handler into stages:
- Enumerate any strings if desired
- Convert any additional properties specifically
- Reduce any single item arrays to a single value if needed

Two test utility methods `expectValid` and `expectInvalid` were added to make the validate tests more legible.  (We should refactor other tests later).

### DIFF
--- a/test/operator-convert.spec.ts
+++ b/test/operator-convert.spec.ts
@@ -3,6 +3,7 @@ import 'mocha';
 
 import { OperatorFactory } from '../src/factory';
 import { ConvertOperator } from '../src/operators';
+import { expectInvalid, expectValid } from './utils/mocha';
 
 const item = {
   quantity: '10',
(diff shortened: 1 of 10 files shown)


test/operator-convert.spec.ts
```diff
@@ -21,40 +22,33 @@ const item = {
...
+    });
+    expectValid({ name: 'convert', properties: 'price,tax', type: 'real' });
+    expectValid({ name: 'convert', properties: ['price', 'tax'], type: 'real' });
+
+    expectInvalid({ name: 'convert' }, 'properties or enumerate or both is needed');
+    expectInvalid({ name: 'convert', properties: 'price' }, 'must also specify type');
+    expectInvalid({ name: 'convert', type: 'real' }, 'must specify properties');
+    expectInvalid({ name: 'convert', enumerate: 'true' }, 'enumerate should be a bool');
+    expectInvalid({ name: 'convert', properties: ['price', 'tax'], type: 'array' }, 'array is not valid');
+    expectInvalid({
```
TrevorFSmith, line 46: Are there conflicting combinations of properties that we should error or warn about?

TrevorFSmith (approved):
LGTM.

I left a comment asking whether the operator should error or warn when incorrectly configured, but that's nit that you can ignore or address at will.

test/operator-convert.spec.ts
van-fs: I _think_ I have them all accounted for.  The one gray area is probably `force` with `enumerate`.  I'll add a card to track.

van-fs:
Validation of rules results in throwing `Error`, which in the `Observer` gets handled as `logger.error`.  I've started to use `warn` more so as an indicator that something unexpected occurred but a total failure did not occur.  For example, failing to register a rule is an error.  Being unable to convert a property is a warning because as messy as data layers are, it's possible to convert successfully on one page but incorrectly on another.  I think it's helpful to dial back this level of feedback so that errors are surfacing more widespread, total failures while warnings could be sporadic or partial failures.

PR #117, opened 2021-02-18
DLO observation messages

TrevorFSmith:
Add a bit more error catching to DLO initialization.
Added timing logs for DLO initialization.

### DIFF
--- a/src/embed/init.ts
+++ b/src/embed/init.ts
@@ -70,10 +70,10 @@ function _dlo_collectRules(): any[] {
         results.push(rule);
       });
     });
-    Logger.getInstance().info(`DLO rules processing time (ms): ${startTime - Date.now()}`);
+    Logger.getInstance().record('DLO rule processing time', { numericValue: startTime - Date.now() });
     return results;
   } catch (err) {
-    Logger.getInstance().error(`Error collecting DLO rules: ${err}`);
+    Logger.getInstance().error(LogMessageType.RuleRegistrationError, { reason: `Error: ${err}` });
     return [];
   }
 }
@@ -105,9 +105,9 @@ function _dlo_initializeFromWindow() {
       urlValidator: win._dlo_urlValidator || undefined,
       rules,
     });
-    Logger.getInstance().info(`DLO init (ms): ${startTime - Date.now()}`);
+    Logger.getInstance().record('DLO initialization time', { numericValue: startTime - Date.now() });
   } catch (err) {
-    Logger.getInstance().error(`Error initializing DLO: ${err}`);
+    Logger.getInstance().error(LogMessageType.ObserverInitializationError, { reason: `Error: ${err}` });
   }
 }
 
--- a/src/embed/init.ts
+++ b/src/embed/init.ts
@@ -53,49 +53,62 @@ window['_dlo_rulesFromOpsTeam'] = [
 */
 
 function _dlo_collectRules(): any[] {
-  const results: any[] = [];
-  Object.getOwnPropertyNames(window).forEach((propName) => {
-    if (propName.startsWith('_dlo_rules') === false) return;
+  try {
+    const startTime = Date.now();
+    const results: any[] = [];
+    Object.getOwnPropertyNames(window).forEach((propName) => {
+      if (propName.startsWith('_dlo_rules') === false) return;
 
-    const prop = (window as { [key: string]: any })[propName];
-    if (Array.isArray(prop) === false) {
-      Logger.getInstance().warn(LogMessageType.RuleInvalid,
-        { property: prop, reason: 'Rules list must be an array' });
-      return;
-    }
+      const prop = (window as { [key: string]: any })[propName];
+      if (Array.isArray(prop) === false) {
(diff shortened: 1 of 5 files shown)


TrevorFSmith:
I've taken the approach of logging timing information as well as catching and logging errors during the embed initialization process.

@van-fs Could you give this a look and see if you're happy with how these messages are structured?

The only thing left for this card is to add logic that watches for event emission rates above some frequency, like say 1000/second and then log that if it occurs. I wonder, though, whether this is something better done with queries on the back end.

src/embed/init.ts
```diff
@@ -53,49 +53,62 @@ window['_dlo_rulesFromOpsTeam'] = [
...
+      prop.forEach((rule: any) => {
+        results.push(rule);
+      });
     });
-  });
-  return results;
+    Logger.getInstance().info(`DLO rules processing time (ms): ${startTime - Date.now()}`);
+    return results;
+  } catch (err) {
+    Logger.getInstance().error(`Error collecting DLO rules: ${err}`);
```
van-fs, line 76: I'm wondering if you can just re-use `LogMessageType.RuleRegistrationError`?  If you need something more particular to this failure, you could add a new field to `LogMessageType`.  The err itself could be passed on as doing this `Logger.getInstance().error(LogMessageType.RuleRegistrationError, { reason: err.message });`

src/embed/init.ts
```diff
@@ -53,49 +53,62 @@ window['_dlo_rulesFromOpsTeam'] = [
...
+        return;
+      }
+
+      prop.forEach((rule: any) => {
+        results.push(rule);
+      });
     });
-  });
-  return results;
+    Logger.getInstance().info(`DLO rules processing time (ms): ${startTime - Date.now()}`);
```
van-fs, line 73: Could we use `debug` versus `info`?  There's some new interest in reducing the amount of ingested events and `debug` would give us a way to keep the feature but not default to the more costly state.

Also, you can pass a `LogContext` as the second arg.  I'm wondering if it makes sense to simply add `timing: number` to `LogContext`.  That way you can make it so we can search greater than or less than for timing entries.

src/embed/init.ts
```diff
@@ -53,49 +53,62 @@ window['_dlo_rulesFromOpsTeam'] = [
...
+      beforeDestination: win._dlo_beforeDestination || undefined,
+      logLevel: win._dlo_logLevel || undefined,
+      previewMode: win._dlo_previewMode === true,
+      previewDestination: win._dlo_previewDestination || undefined,
+      readOnLoad: win._dlo_readOnLoad === true,
+      validateRules: win._dlo_validateRules === true,
+      urlValidator: win._dlo_urlValidator || undefined,
+      rules,
+    });
+    Logger.getInstance().info(`DLO init (ms): ${startTime - Date.now()}`);
```
van-fs, line 108: Previous comments apply here: debug, timing context, externalize the string, and pass err as reason.message.

src/observer.ts
```diff
@@ -108,7 +109,9 @@ export class DataLayerObserver {
 
     if (rules) {
       rules.forEach((rule: DataLayerRule) => this.registerRule(rule));
+      Logger.getInstance().info(`DLO rule count: ${rules.length}`);
```
van-fs, line 112: We might also add `count?: number` to `LogContext`.

src/observer.ts
```diff
@@ -108,7 +109,9 @@ export class DataLayerObserver {
...
     }
+    Logger.getInstance().info(`DLO constructor time (ms): ${startTime - Date.now()}`);
```
van-fs, line 114: Maybe a new `LogMessageType.Initialization` field and passing the `timing` in the context.

van-fs:
Overall looks good.  I was advocating using `debug` versus `info` since timing is a more granular level of information.  I suppose that may inhibit observability if the intent is for "us" to be able to look into some of this data.  But if the desire is for an end-user admin to review, I'd prefer `debug` since we default to logging at `error`.

van-fs:
Why don't we also shift the the stats debugger from a console log to something more structured like the rest of the perf data in this PR.  So we'd swap this [line](https://github.com/fullstorydev/fullstory-data-layer-observer/blob/edfd3a2a7ce30e6b9ab6de03564687b5d6de5d9d/src/handler.ts#L127) for a standard `Logger.debug` call.  We'll have to create some new fields in `LogContext`, but that would allow us to better track data consumption.  I think we can then remove the `runDebugger` function.

TrevorFSmith:
Ok, @van-fs this is ready for another review. I made changes based on our conversation about how to handle statistics. The stats path through the logger is stubbed out awaiting future changes outside of DLO.

van-fs:
LGTM.  I have a few bigger questions that I don't think get answered until the `stats` API is live.  For example, I'd want to `record` whenever we have and `error` but only for the FullStoryAppender.  I think I'll just cross that bridge when I get there.  @TrevorFSmith this is still in draft, did you want to mark it ready for review and merge?

van-fs (approved):

PR #119, opened 2021-03-02
add suffix bug fix and bump version

van-fs:
This PR fixes a bug where an `FS` API includes the `source` param.  Previously, the suffix operator expected that the object to suffix was last in the list of args.  Adding the `source` param changes that assumption. The suffix operator succeeds but sends malformed data to FullStory.  This results in FullStory rejecting the event with [Expected Atom](https://help.fullstory.com/hc/en-us/articles/360020624394-Troubleshooting-Recording-Client-API-FS-event-FS-setUserVars-and-FS-identify-errors#expectedAtom) errors.

### DIFF
--- a/src/operators/suffix.ts
+++ b/src/operators/suffix.ts
@@ -160,7 +160,13 @@ export class SuffixOperator implements Operator {
   }
 
   handleData(data: any[]): any[] | null {
-    const index = this.index >= 0 ? this.index : data.length + this.index;
+    let index = this.index >= 0 ? this.index : data.length + this.index;
+
+    // check if the `source` param was included and if so decrement the index
+    if (typeof data[index] === 'string') {
+      index -= 1;
+    }
+
     const suffixedData = data;
     suffixedData[index] = this.mapToSuffix(suffixedData[index]);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 1.6.1
+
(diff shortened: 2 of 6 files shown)


TrevorFSmith (approved):

src/operators/suffix.ts
```diff
@@ -160,7 +160,13 @@ export class SuffixOperator implements Operator {
   }
 
   handleData(data: any[]): any[] | null {
-    const index = this.index >= 0 ? this.index : data.length + this.index;
+    let index = this.index >= 0 ? this.index : data.length + this.index;
+
+    // check if the `source` param was included and if so decrement the index
```
patrick-fs, line 165: I know this is already merged, but how is the source parameter provided?

van-fs: It's coming in from the `beforeDestination` setting.  So what's happening is that every rule is automatically getting the `suffix` operator run on it before the destination.  Where the real problem lies is souq, which is now adding the source parameter.  But since the suffix operated had defaulted to looking at the last arg (FS.event and FS.setUserVar both have objects as the last arg), this is a problem.  This "fix" essentially corrects what is more accurately a bug in souq, but it would guard against a rule write that for whatever reason also added the source param.

patrick-fs: Got - I leveled up a little :)

Would this have had the same desired outcome if the logic were

```
if (typeof data[index] !== 'object') {
      index -= 1;
    }
```

?

van-fs: It would.  It's also an edge case, but we'd have to delineate between a POJO and an array since the typeof for an array returns `object`.  If we want to adjust the check a bit, I could make that change.

PR #120, opened 2021-03-03
Initialize the logger early in the embed script 

TrevorFSmith:
In embed, the Logger needs to be initialized with the custom appended (`fullstory`) so that stats and early failures are correctly appended.


van-fs (approved):
LGTM.

PR #121, opened 2021-03-12
Cleanup mocha tests

van-fs:
I performed a bit of spring cleaning on some of the mocha tests:
- Split out vendor-specific rules from a single spec into vendor-specific specs.
- Added a few convenience methods to deal with globals like `FS`.
- Over time there was a fair amount of entropy between how we tested rules.  I added a `expectEqual` to more appropriately perform deep equality since sometimes we used `eq` and other times `eql`.
- We often checked equality of a destination payload with the data layer that created N lines of `expect`. I added a `expectMatch` and `expectUndefined` to make that more succinct.

The tests are the same, but the chai assertions have all been updated to use the respective helper functions.


test/utils/mocha.ts
```diff
@@ -1,11 +1,79 @@
...
+/**
+ * A setup function that inits the global object (e.g. window) with desired expandos.
+ * Each expando will be deep copied to prevent cross-contamination between tests.
+ * Additionally, a FullStory mock will also be added to the `FS` expando.
+ * @param tuples List of expando tuples (e.g. ['FS', object])
+ */
+export function setupGlobals(tuples: [string, any][]) {
+  tuples.forEach((tuple) => {
+    expect(tuple[0]).to.be.ok;
+    expect(tuple[1]).to.be.ok;
```
patrick-fs, line 33: Wouldn't this be an overflow issue if only one tuple were passed into this function?

patrick-fs: It looks like you could define this as a type with shape:

beforeEach(() => setupGlobals([
    ['digitalData', basicDigitalData],
    ['_dlo_rules', rules],
  ]));

```
type dataLayerName =  "digitalData" | "s" | "dataLayer" | "utag";
type ruleName = "_dlo_rules";
type dataLayerDefinition = [dataLayerName, any];
type ruleDefinition = [ruleName, any];
setupGlobals(datalayer: dataLayerDefinition, rule: ruleDefinition)
```

I'm handing code so maybe there's an issue with this, but it could spare a potential bug down the line and is a maybe bit more descriptive.

test/utils/mocha.ts
```diff
@@ -1,11 +1,79 @@
...
+import { DataLayerObserver, DataLayerConfig, DataLayerRule } from '../../src/observer';
 import MonitorFactory from '../../src/monitor-factory';
 import { BuiltinOptions, OperatorFactory } from '../../src/factory';
 
+/**
+ * Gets or sets a global object. Use this rather than writing `globalThis as any`.
+ * @param key The key corresponding to the global object (e.g. FS in window['FS']).
+ * @param value Optional object that when provided will set the global's value
+ */
+export function global(key: string, value?: any): any {
```
patrick-fs, line 16: it might be better to be more explicit with `getGlobal` and `setGlobal` -  this way the behavior doesn't change depending on whether a parameter is present.

test/utils/mocha.ts
van-fs: I made a few adjustments to the comments and function.  The main goal of the function was to reduce the overhead of setting up the `window`.  Since the `window` can take `any` value as an expando, that was the rationale to use less stringent type checking.  The function block doesn't make any type assumptions, it simply creates and assigns the expando (most importantly deep copying the provided value).  I did, however, beef up some of the expects inside the function block to be more informative on usage.

All that said, I'm wondering if this "setup" function should just add _all_ supported data layers.  I've skewed more to testing rulesets in isolation, but there's not really much overlap due to the different vendor's global objects.  What this approach does not address is future situations where I might want to set an arbitrary expando like `_dlo_destination`.  I could blend having the data layers auto-created and an optional that allows you to pass in additional key:value pairs for additional expandos.

test/rules-tealium-fullstory.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    ExpectObserver.getInstance().create({
+      // ensure this rule never reads on load
+      rules: [{ ...expectRule('fs-tealium-event'), readOnLoad: false }],
+    });
+
+    // NOTE this is a valid property to monitor
+    expectGlobal('utag').data.product_id = ['PROD789'];
+
+    // check the assignment
+    setTimeout(() => {
```
patrick-fs, line 49: why run these assertions in a `setTimeout`?

test/utils/mocha.ts
patrick-fs: I was thinking the explicit types would add more clarity around the how data layers are named. I don't have a strong opinion, and I think your current implementation is fine.

> I'm wondering if this "setup" function should just add all supported data layers. I've skewed more to testing rulesets in isolation, but there's not really much overlap due to the different vendor's global objects.

I think you should set them up individually for each datalayer vendor test (as you are currently). This gets to my point around clarity of which data layer definition is tied to which vendor.

patrick-fs (approved):
LGTM - I had a question just out of curiosity.

test/rules-tealium-fullstory.spec.ts
van-fs: The `DataHandler` debounces the incoming events so that multiple assignments to an object's properties don't fire N `FS.events`.  For example, say the `cart` was sequentially assigned `total = 12.99` and `tax = 1.99`.  It waits for a timeout period when no new assignments have occurred and then sends the final `cart` object with the two updated properties.

But this is a good example of where this has been a copy/paste job across a few tests that should be codified in a util function.  I created a card to track.

PR #126, opened 2021-05-14
Fix data layer side effect from flatten operator

van-fs:
This PR fixes an issue where the flatten operator unintentionally updates the data layer.  The issue was due to assignment of new data to a shallow reference.  Flatten is an older operator, and it did not have the convention seen in more recent operators:

```
const clone = data.slice();
clone.splice(this.index, 1, flattened);
return clone;
``` 

I added additional test cases to the `flatten.spec.ts` and carried over the same test to two other specs (`convert.spec.ts` and `rename.spec.ts`) just to confirm they also did not have a similar bug.  No problems were found with these operators, and the convert operator has numerous checks for data layer side effects.

### DIFF
--- a/src/operators/flatten.ts
+++ b/src/operators/flatten.ts
@@ -49,10 +49,11 @@ export class FlattenOperator implements Operator {
   }
 
   handleData(data: any[]): any[] | null {
-    const flattenedData = data;
-    flattenedData[this.index] = this.flattenHelper(flattenedData[this.index]);
+    const flattened = this.flattenHelper(data[this.index]);
 
-    return flattenedData;
+    const clone = data.slice();
+    clone.splice(this.index, 1, flattened);
+    return clone;
   }
 
   validate() {
--- a/src/operators/flatten.ts
+++ b/src/operators/flatten.ts
@@ -1,4 +1,7 @@
-import { Operator, OperatorOptions, OperatorValidator } from '../operator';
+/* eslint no-param-reassign: ["error", { "props": false }] */
+import {
+  Operator, OperatorOptions, OperatorValidator, safeUpdate,
+} from '../operator';
 
 export interface FlattenOperatorOptions extends OperatorOptions {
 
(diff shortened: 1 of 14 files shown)


test/operator-rename.spec.ts
```diff
@@ -39,6 +39,24 @@ describe('rename operator unit tests', () => {
...
+          data: 'datum',
+        },
+      },
+    );
+
+    const customList = [{ event: 'test', data: { foo: 'bar' } }];
+    const [flatData] = operator.handleData(customList)!;
+    expect(customList[0].data).to.not.be.undefined;
+    expect(flatData.data).to.be.undefined;
+    expect(flatData.datum).to.be.not.undefined;
```
sabrina-li, line 57: nit: flatData -> renamedData?

src/operators/flatten.ts
```diff
@@ -49,10 +49,11 @@ export class FlattenOperator implements Operator {
   }
 
   handleData(data: any[]): any[] | null {
-    const flattenedData = data;
-    flattenedData[this.index] = this.flattenHelper(flattenedData[this.index]);
+    const flattened = this.flattenHelper(data[this.index]);
```
sabrina-li, line 52: maybe note that `flattenHelper` is pure (no side effects) ?

sabrina-li (approved):
LGTM! left some nits

src/operators/flatten.ts
sabrina-li: actually the helper function isn't "pure" but the way it's called isn't mutating the data.

van-fs: Yes, I was just writing the comments now.  The fact that the `node` param is not passed has it create a new object under the hood.  I'm going to tweak this code a bit so that's more obvious.

van-fs:
@sabrina-li Could you do another review.  I tweaked the `flattenHelper` because the optional `node` argument felt like a footgun.  I also moved the best practice code of "copying" the data layer before returning any transformed data into a `safeUpdate` function.  (I'm open to other names for it.) Having it enforces more consistency, and I've updated any operators that transform/update to have comments signaling they do so and use the `safeUpdate` function.

sabrina-li:
LGTM!

PR #127, opened 2021-05-17
Undefined empty events

van-fs:
Normally, a data layer event where the observed changed was `undefined` triggered a warning log message.  This was designed to signal that a possible DLO bug was present.  It appears that there may be data layers where `undefined` is assigned to an observed property as a way to "clear" variables in the data layer.  This may be done at a high volume in which case, `warn` is undesirable because it could cause rate limiting.  This PR moves the log level from `warn` to `debug` to allow this scenario while still retaining the ability to know when an undefined assignment is occurring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 1.6.4
+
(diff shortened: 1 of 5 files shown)


sabrina-li (approved):
LGTM!

PR #128, opened 2021-05-26
Suffix Corrections

van-fs:
The new page vars API was not allowing us to set the `pageName` correctly because we by default auto-suffix all properties.  When the property was set to `pageName_str`, FullStory did not correctly infer the page name because it is specifically looking for `pageName`.  Adjustments were made to `displayName` and `email` as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Sensitive, private, and confidential information should never be added to a data
 
 DLO is a JavaScript asset that is included on a web page.  FullStory hosts versions of DLO on our CDN.  Versioned releases have the naming convention `<version>.js`, and the most recent version is named `latest.js`:
 
-- https://edge.fullstory.com/datalayer/v1/1.6.4.js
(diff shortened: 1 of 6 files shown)


sabrina-li (approved):
that's an interesting convention for pagevars, I guess it's fine to just hard code these for now since they are the only special ones. Would this affect "generic users" where the destination is not FS, I guess the chances of that is not very hight so this should be fine too.
Unrelated to this pr, might be good to look for both suffixed and non-suffixed in the pagevars.
but overall LGTM

van-fs:
Right, the implementation forces some behavior - though there is diminishing probability it leads to an issue:

- You have to first want to suffix (i.e. configure to use the `suffix` operator, which is specific to FullStory)
- These properties would also have to exist in the root object (if the properties are nested, we don't convert)
- You're unable to undo this behavior through custom rules and changing the DLO configuration expandos

PR #129, opened 2021-05-27
Add google rulesets

van-fs:
For a while we've had [examples](https://github.com/fullstorydev/fullstory-data-layer-observer/tree/main/examples) of possible rules, and many examples were used in the production integration for DLO.  Over time it became less clear what were the actual rules used in the integration versus possibilities because they were copied directly into Go code.  (I really really want to get away from this choice, but that's a PR for another day.)  

I've added a rulesets folder with the intent to begin to re-define which rules are being used in production (or act as THE recommended starting point for all consumers) along with references to vendor documentation and test cases.  Something to note, the rules are essentially a copy/paste exercise with some code cleanup in the specs.  There's no effect to production behavior.

Once new rulesets land fully, we should decide what to do with examples (are they needed?).  And then we'll need to reconcile rules used in Souq with these.


sabrina-li (approved):
LGTM

Issue #108, opened 2020-11-18
Add more than one beforeDestination operator


van-fs:
The `_dlo_beforeDestination` currently accepts only one operator.  Ideally, this could be many as multiple operations might need to be performed prior to the destination.  I propose that we change to also support a list of operators.  This will allow us to also set the `source` param for FS.* destination automatically.

van-fs:
Cleaning up; this feature is now included.

PR #132, opened 2021-06-07
Allow more than one operator for beforeDestination

van-fs:
This feature has been on the backlog for a while re: #108 .  It's become more clear that there's a need for multiple, required operators to support FullStory use cases: suffixing, adding the `source` param, and now auto-converting strings to numbers using the `enumerate: true` feature of the `convert` function.

So what I've done is updated the `beforeDestination` feature to take a list of always-executed operators prior to calling the `destination` function.  This change will support more concise configuration such as `{window['_dlo_beforeDestination'] = [{ name: 'convert', enumerate: true }, { name: 'suffix' },  { name: 'insert', value: 'dlo', index: -1 }]`.

For completeness, it's worth doing some level of performance analysis to see what an "always on" auto-conversion yields.  I'll do that prior to merging but wanted to show where things are headed.

### DIFF
--- a/src/operators/convert.ts
+++ b/src/operators/convert.ts
@@ -210,14 +210,17 @@ export class ConvertOperator implements Operator {
         : !Number.isNaN((newValue as Date).getTime());
     }
 
-    // log warning and reset to the original value
+    // Note `debug` level is used because `enumerate` may always be done `beforeDestination`, which could generate
+    // a lot of false positives
+    // log debug and reset to the original value
     if (!verified) {
-      Logger.getInstance().warn(LogMessageType.OperatorError, {
+      newMap[property] = oldValue; // eslint-disable-line no-param-reassign
+
+      Logger.getInstance().debug(LogMessageType.OperatorError, {
         operator: 'convert',
         property: property.toString(),
         reason: `Failed to convert to ${type} for value ${oldValue}`,
       });
-      newMap[property] = oldValue; // eslint-disable-line no-param-reassign
     }
   }
 }
--- a/src/operators/convert.ts
+++ b/src/operators/convert.ts
@@ -83,6 +83,8 @@ export class ConvertOperator implements Operator {
   handleData(data: any[]): any[] | null {
     // NOTE this operator transforms data - be absolutely sure there are no side effects to the data layer!
 
+    const index = this.index >= 0 ? this.index : data.length + this.index;
+
     let { properties } = this.options;
     const {
       enumerate, force, preserveArray, type,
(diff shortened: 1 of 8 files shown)


test/observer.spec.ts
```diff
@@ -319,7 +319,8 @@ describe('DataLayerObserver unit tests', () => {
   it('it should register and call an operator before the destination', () => {
```
patrick-fs, line 319: nit: maybe break this into two test cases: one for a single operator and one for multiple operators before destination?

Up to you.

test/observer.spec.ts
```diff
@@ -333,12 +334,42 @@ describe('DataLayerObserver unit tests', () => {
...
 
     ExpectObserver.getInstance().cleanup(observer);
+
+    expectNoCalls(globalMock.console, 'log');
+
+    // test multiple operators
+    observer = ExpectObserver.getInstance().create({
+      beforeDestination: [
+        { name: 'toUpper' },
+        { name: 'suffix' }, // NOTE suffix is a built-in operator
```
patrick-fs, line 350: Do you also need to explicitly test this, or is it covered implicitly in `expect(category.primaryCategory_str)`

test/observer.spec.ts
van-fs: It’s a good idea; will update.

patrick-fs (approved):
LGTM - I had a couple of questions

test/observer.spec.ts
van-fs: The expect on ‘ category.primaryCategory_str’ is checking that the suffix operator actually suffixes the property (ie a suffixed property exists and it’s set to the corresponding value).  Then there’s this more general ExpectObserver that checks the observer is created correctly.  We def need to check the property in the emitted object from the operator chain, but there’s an opportunity to clean up the test harnesses if it’s not immediately intuitive.

src/operators/convert.ts
```diff
@@ -210,14 +212,17 @@ export class ConvertOperator implements Operator {
         : !Number.isNaN((newValue as Date).getTime());
     }
 
-    // log warning and reset to the original value
+    // Note `debug` level is used because `enumerate` may always be done `beforeDestination`, which could generate
+    // a lot of false positives
+    // log debug and reset to the original value
     if (!verified) {
-      Logger.getInstance().warn(LogMessageType.OperatorError, {
+      newMap[property] = oldValue; // eslint-disable-line no-param-reassign
```
van-fs, line 219: Note I moved the assignment above the Logger in case something in it could've prevented that from occurring.  (I could've also try catched the Logger, but figured this was sufficient with less code.)

src/operators/convert.ts
```diff
@@ -83,6 +83,8 @@ export class ConvertOperator implements Operator {
   handleData(data: any[]): any[] | null {
     // NOTE this operator transforms data - be absolutely sure there are no side effects to the data layer!
 
+    const index = this.index >= 0 ? this.index : data.length + this.index;
```
van-fs, line 86: This is really the main change.  The lines below have find/replace on `this.index` with `index`.

van-fs:
@patrick-fs Can you do another review when you can?  I found that the primary use-case that started this PR, using `{ name: 'convert', enumerate: true }`, didn't work.  I had to update `convert` so that it would look at the end of the arguments list.  Previously, convert just assumed you wanted a positive index (e.g. convert the first `0` item of a list).  Having it read from the end of a list allows us to support `FS.event(eventName, payload)` and `FS.setUserVars(payload)` with the `{ name: 'convert', enumerate: true, index: -1 }` operator.  I updated the README to reflect this as a best practice.

test/observer.spec.ts
patrick-fs: I think my comment was about clarity in the tests - maybe just a comment above `expect(category.primaryCategory_str).to.eq(` that states the expectation also covers the suffix operator will indicate that expected suffixing behavior is covered too.

patrick-fs:
> @patrick-fs Can you do another review when you can? I found that the primary use-case that started this PR, using `{ name: 'convert', enumerate: true }`, didn't work. I had to update `convert` so that it would look at the end of the arguments list. Previously, convert just assumed you wanted a positive index (e.g. convert the first `0` item of a list). Having it read from the end of a list allows us to support `FS.event(eventName, payload)` and `FS.setUserVars(payload)` with the `{ name: 'convert', enumerate: true, index: -1 }` operator. I updated the README to reflect this as a best practice.

Hey Van, why does the convert operation have to start from the last entry in the list of properties? The behavior outlined in the convert operator doc (https://github.com/fullstorydev/fullstory-data-layer-observer/blob/main/docs/operator_convert.md#automatically-convert-all-numeric-strings) makes sense to me, but I think I'm not understanding what the 'index' does.

van-fs:
@patrick-fs Understandable.  So `destination` is a function.  To be flexible, DLO uses [Function.apply](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) to pass the arguments from the operator pipeline to that `destination` function like `FS.event` or `FS.setUserVars`.  An object that changes in the data layer, let's call it `obj`, get dispatched as an event broadcasting the current state of `obj`.  Each operator receives a list, and the very first operator receives `[obj]` with the data emitted from the data layer.  The reason it receives a list (and not just the single object itself) is because DLO also shims functions.  So a call to `track(valueA, valueB)` passes through the operator chain as `[valueA, valueB]`.  Thus you have to tell each operator which index in the list you want to operate on.

Most often, we're operating on an object coming out of the data layer, and the list is just a single element (the `obj`).  For this reason, all operators default to operating on the first element in the list and `index` may be omitted.  `index` is a common design for all operators - not just `convert`, and this design is what allows us to take this list and pass it as `arguments` to the `destination` function through `apply`.

The method signature of FullStory APIs has the object payload as the last (public) argument:
`FS.event(eventName, obj)`
`FS.setUserVars(obj)`
`FS.identify(uid, obj)`
`FS.setVars(page, obj)`

The commonality is that the object we want to `{ name: 'convert', enumerate: true }` is the last object in the list of arguments to the function.  So `beforeDestination`'s `convert` needs to look at the last object in the list: for example  FS.event's `[eventName, obj]` or FS.setUserVar's `[obj]`.

Side note is that every operator should really allow reading from the start of a list or the end of the list.  Currently, suffix and convert have been the only ones that truly need this ability, but it's a nice to have feature for every operator.

PR #134, opened 2021-06-25
Remove tsconfig.json comments

van-fs:
Internal build warns because comments exist in JSON.  Removing comments.


sabrina-li (approved):
LGTM

PR #135, opened 2021-06-26
Increased logging visibility

van-fs:
I've seen an uptick in error logs that indicate the `name` property is missing from operator options.  I can't understand how that's happening so I've added additional logging to print the raw operator that gets passed into configuration.


sabrina-li (approved):
LGTM

PR #136, opened 2021-07-02
Fix read on load for rules

van-fs:
Looks like the logic for overriding the global config for `readOnLoad` is incorrect.  If DLO is globally configured as `readOnLoad:true`, it's not possible to make a rule that had `readOnLoad:false`.  This now checks if a rule has `readOnLoad` defined and sets the behavior accordingly.


sabrina-li (approved):
LGTM

PR #137, opened 2021-07-09
Enumerate shoudn't convert empty strings

van-fs:
Historically the `ConvertOperator` allowed you (end-user) to specify what to convert.  The default behavior of `convert` on an empty string to a `real` is that the resulting value is `0`.  After introducing `enumerate` where `convert` tries to auto-convert for convenience, this has what appears to be an unwanted side effect.  If the properties selected from an object in the data layer are broad (e.g. all eVars), many times those get "cleared" by setting the existing value to empty string.  DLO then converts it to a 0, which is probably unexpected and has the second drawback of creating `eVarN_str` and `eVarN_real` in custom event properties.  So it feels best to just leave empty strings as-is for enumeration specifically.

Obscure technical detail: while the code change now checks if a single value is an empty string, the code to enumerate strings in a list is unchanged (i.e. we don't have any check for empty strings in a list).  The thinking is this, `verifyConversion` already checks to see if every element in a list is the same type.  If they don't, we assume conversion didn't work and reset the list to its original value.  Because FullStory doesn't allow mixed lists (what suffix would that be?), the conversion is all or none.  So a list of `['', '1.99']` would expectedly become `[0, 1.99]` but `['', 'foo']` gets converted to `[0, 'foo']` and then we `verifyConversion` which fails so its set back to `['', 'foo']`.

### DIFF
--- a/src/operators/convert.ts
+++ b/src/operators/convert.ts
@@ -46,6 +46,9 @@ export class ConvertOperator implements Operator {
       case 'date': return new Date(value);
       case 'int':
       case 'real':
+        // NOTE be careful of trying to convert an empty string, which will become 0
+        // a guard exists in `enumerate` but `convert` assumes you intend the conversion
+        // and the ternary allows converting a boolean for example to 0
         return !value ? 0 : ConvertOperator.enumerate(value);
       case 'string':
         switch (typeof value) {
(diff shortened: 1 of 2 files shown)


patrick-fs (approved):
LGTM - I just left one `nit` comment

PR #138, opened 2021-07-09
Guard emitting the same value

van-fs:
Some object-based data layers appear to be re-assigning the same value to properties very frequently.  Conceptually that is problematic because it produces a change event that triggers the operator chain, which would send an event to the destination.  This scenario has been less noticed because property setters are debounced higher in the call stack by the `DataHandler`.  So a no-op, same-value reassignment occurs within the debounce window of a different property's assignment of a different value.  When the debounce time limit is reached, the two properties get collected, which looks like a valid change event because one property's did change.  As the list of properties under observation in the object grows, it appears the incidence of these "same value" assignments can occur.  For super active/messy? data layers, it increases the possibility that change events are emitted for data that's already been seen/processed.  I have another PR on allowing that debounce window to be extended to further reduce this scenario.

Also, it's probable that there's some legitimate scenario where you'd want to receive an event for a re-assignment, but I feel like that's a lesser problem to an over-active/messy data layer that's susceptible to rate limiting.  Open to hearing counter points.

### DIFF
--- a/src/monitor-shim.ts
+++ b/src/monitor-shim.ts
@@ -43,8 +43,15 @@ export default class ShimMonitor extends Monitor {
       enumerable: this.enumerable,
       get: () => this.state,
       set: (value: any) => {
+        // checks if the incoming value matches the existing value
+        const hasChanged = this.state !== value;
+
         this.state = value;
-        this.emit(value);
+
+        // object-based data layers will often re-assign the same value, which causes over emitting
+        if (hasChanged) {
+          this.emit(value);
+        }
       },
     });
   }
--- a/test/shim-monitor.spec.ts
+++ b/test/shim-monitor.spec.ts
@@ -66,6 +66,35 @@ describe('ShimMonitor unit tests', () => {
     globalMock.digitalData.cart.cartID = 'cart-5678';
   });
 
+  it('it should not emit an on change value if it matches the existing value', () => {
+    const path = 'digitalData.transaction';
+
+    // this test is a bit different because we have to create a listener where the
+    // Nth invocation is expected not to happen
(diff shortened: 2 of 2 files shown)


src/monitor-shim.ts
```diff
@@ -43,8 +43,15 @@ export default class ShimMonitor extends Monitor {
       enumerable: this.enumerable,
       get: () => this.state,
       set: (value: any) => {
+        // checks if the incoming value matches the existing value
+        const hasChanged = this.state !== value;
```
sabrina-li, line 47: Just for my own kownledge: `this.state` is the current value of the property, and we are doing a shallow compare here to only emit the change event when they are not the same, right? 
When the property is an object, do we recursively shim, or is that even a valid configuration?

sabrina-li (approved):
LGTM, I left a question for my own learning

src/monitor-shim.ts
van-fs: Yes.  This is really to guard against primitive assignments that are the same.  For objects, we don't do a deep comparison to figure out if the content itself is different.  It feels like the performance cost of doing that probably outweighs any downstream benefit.

DLO does recursively shim. There's some technical nuance that probably needs revisiting, but if the target is `digitalData.cart`, we shim all properties in `cart` and we _should_ be shimming all properties in child objects.  This monitor code looks at the property specifically for changes, and if it happens it's emitted.  So if you did:

```
digitalData.cart.cartID = '1234';
digitalData.cart.total = 29.99;
```

That would be two distinct changes that we'd rely on debouncing to help smooth sequential changes into a single "event".

PR #139, opened 2021-07-11
Add configurable debounce time

van-fs:
Recently, I've found that very active data layers (object-based in particular) can over-emit events.  I found a few areas of improvement:
- adding a config option to set the debounce time allowing some rules more time for the data layer to cool down
- moving the `this.timeoutId = null;` to unset the debounce flag; I saw that for a really really data layer I saw code that had its timeout cleared still executed.  Maybe some issues with the scope?  Open to theories on this change.

### DIFF
--- a/test/observer.spec.ts
+++ b/test/observer.spec.ts
@@ -488,6 +488,9 @@ describe('DataLayerObserver unit tests', () => {
       }],
     }, true);
 
+    // there are some technical nuances to when/why to use debouncing
+    // see https://github.com/fullstorydev/fullstory-data-layer-observer/pull/139#discussion_r667939622
+
     // #1 trigger the first change event
     expectGlobal('digitalData').cart.cartID = 'abc';
 
--- a/test/observer.spec.ts
+++ b/test/observer.spec.ts
@@ -9,7 +9,7 @@ import {
 import Console from './mocks/console';
 import FullStory from './mocks/fullstory-recording';
 import {
-  expectParams, expectNoCalls, expectCall, ExpectObserver,
+  expectParams, expectNoCalls, expectCall, ExpectObserver, expectGlobal, expectEqual,
 } from './utils/mocha';
 import { Operator, OperatorOptions } from '../src/operator';
 import { LogEvent, LogLevel } from '../src/utils/logger';
@@ -199,7 +199,7 @@ describe('DataLayerObserver unit tests', () => {
     ExpectObserver.getInstance().cleanup(observer);
   });
 
-  it('invalid operators should remove a handler', () => {
+  it('invalid operators should remove a handler', (done) => {
     const observer = ExpectObserver.getInstance().create({
       validateRules: true,
       rules: [{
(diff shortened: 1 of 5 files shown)


test/observer.spec.ts
```diff
@@ -465,7 +468,49 @@ describe('DataLayerObserver unit tests', () => {
...
+    // #3 delay a subsequent change such that it occurs outside the debounce window
+    setTimeout(() => {
+      expectGlobal('digitalData').cart.cartID = 'xyz';
+    }, 20);
+
+    // check the final assignment and that two events were queued to the destination (#2 and #3)
+    setTimeout(() => {
+      // check that two events were queued (#1 and #2 coalesced to a single event and #3 was its own event)
+      expectEqual(changes.length, 2);
+      expectEqual('def', changes[0][0].cartID);
```
patrick-fs, line 506: if the cart id value update to 'def' occurred within the debounce window, wouldn't we expect the `cartID` value to be 'abc' here?

van-fs: Unfortunately, the DLO design is running into a constraint.  One one side, we have rate limiting downstream (FullStory).  The other is upstream changes (data layer - think CEDDL object).  So the debounce window (if increased) allows more property assignments to occur within it before sending the data on to the destination.  There are two situations:

1. Many different properties are changing at once.  A lengthened debounce window allows all/most of them to get assigned before sending data to the destination.  This is the ideal situation because we process in bigger batches to not trigger rate limiting.
2. The other is that within that debounce window, the same property's value is changing to something different.  Then the lengthened debounce window works against us because the first assignment gets "overwritten" during the debounce window.  That situation would need reducing the debounce window, which increases the possibility for rate limiting.

But the configurable option gives us a bit more control over which way we want to lean.  A nuanced issue is that most out of the box DLO rules watch so many properties, it increases situation where we need to extend the debounce window.  Another tactic is to create individual rules that more narrowly observer a set of properties and adjust the debounce time as needed.

As this test is written, `abc` triggers a change and the debounce timer starts.  Before the debounce time limit is reached, `def` gets assigned.  The debounce time limit is reached, and DLO sends the last value it saw as part of the data payload.  The other `xyz` assignment happens after the time limit so it's a new data event.

I'm seeing situation 1 currently.  With all that said, I can create two tests to demonstrate the above.  WDYT?  Does that help explain why `def` would be seen?

patrick-fs: Ah, ok. So this is more of a test that there are 2 change events instead of 3, i.e. 2 destination calls instead of 3. 

I think testing for specific values in the `change` array obscures this, but it does shed light on the nuance you've outlined. Maybe just adding a comment in the test that links to this thread is a good enough way to answer any questions that future code reviewers have about what is being tested.

van-fs: Ugh `change` is poor wording on my part then.  I have a big refactor of the test harness where `change` in this case is an auto-created `destination`.  Hopefully that clarifies the usage of the final data - it's the same as the `destination` API you'd provide in a rule  See the same test from that PR https://github.com/fullstorydev/fullstory-data-layer-observer/blob/fc456c5c460cef391e93b40decc020f532f3023d/test/observer.spec.ts#L361.

PR #141, opened 2021-08-02
Limit number of suffixed props

van-fs:
There was a Clubhouse ticket where a customer had long delays on their site.  This was tracked back to a super-size (120MB) object being added to the data layer.  The object had a reference to the React app, which is why it was so big.  In order to guard against wayward objects in the data layer either impacting site performance or simply overrunning our cardinality quota on `FS.event`, I've put a cap on the `suffix` operator's total property count at `100`.  This was done through an informed analysis of DLO events and is not arbitrary.

The rationale for doing this in the `suffix` operator is because it is always used in `beforeDestination` and has FullStory specific behavior already.  Thus it makes for a good candidate to put a guard that benefits FullStory.

### DIFF
--- a/src/operators/suffix.ts
+++ b/src/operators/suffix.ts
@@ -31,7 +31,7 @@ export enum Suffixes {
 }
 
 export interface SuffixOperatorOptions extends OperatorOptions {
-
+  maxProps?: number;
 }
 
 /**
@@ -41,18 +41,24 @@ export class SuffixOperator implements Operator {
   static specification = {
     index: { required: false, type: ['number'] },
     maxDepth: { required: false, type: ['number'] },
+    maxProps: { required: false, type: ['number'] },
   };
+
+  static readonly DefaultMaxProps = 100;
 
   readonly index: number;
 
   readonly maxDepth: number;
 
+  readonly maxProps: number;
+
   constructor(public options: SuffixOperatorOptions) {
     // NOTE the index is -1 because payloads to FS.event or FS.setUserVars are the last in the list of args
-    const { index = -1, maxDepth = 10 } = options;
+    const { index = -1, maxDepth = 10, maxProps = 100 } = options;
 
     this.index = index;
     this.maxDepth = maxDepth;
+    this.maxProps = maxProps;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/operators/suffix.ts
```diff
@@ -127,15 +133,24 @@ export class SuffixOperator implements Operator {
...
     if (obj === undefined || obj === null) {
       return suffixedObj;
     }
 
+    // count the props to guard against super-sized object being unknowingly added to a data layer
+    // this reduces the likelihood of hitting cardinality but also prevents impacting site performance
+    const numProps = totalProps + Object.getOwnPropertyNames(obj).length;
+
+    if (numProps > this.maxProps) {
+      throw Error(`Number of object properties exceeds the limit (${this.maxProps}); increase maxProps to ${numProps}`);
```
sabrina-li, line 151: curious: does this mean we will abort the pipeline rather than sending partial results?

van-fs: That's a great question.  Currently, this aborts and logs and error ... we _could_ however process the 100 properties and still log.  My thinking was a missing event is much clearer that something is wrong versus partial results.  Thoughts on how you think it should work?

src/operators/suffix.ts
```diff
@@ -41,18 +41,24 @@ export class SuffixOperator implements Operator {
...
   readonly index: number;
 
   readonly maxDepth: number;
 
+  readonly maxProps: number;
+
   constructor(public options: SuffixOperatorOptions) {
     // NOTE the index is -1 because payloads to FS.event or FS.setUserVars are the last in the list of args
-    const { index = -1, maxDepth = 10 } = options;
+    const { index = -1, maxDepth = 10, maxProps = 100 } = options;
```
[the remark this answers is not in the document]
van-fs: Not a nit, good catch. I think I went back and made it a static default when I was creating the test cases.  I've updated to use the static default.

src/operators/suffix.ts
van-fs: The other consideration is this: a `100` event property may actually signal that what's in the data layer is anomalous.  If we do a partial recording of data, then we're ingesting "suspicious" data, which has the end result of using up cardinality.

sabrina-li: Ah okay yeah that makes sense, thanks for the explanation. I think this looks good then.

sabrina-li (approved):
LGTM

PR #143, opened 2021-08-06
Limit maxProps to FullStory 5000 cardinality

van-fs:
Put a ceiling on the suffix operator's `maxProps` feature.  No rush on reviewing this one, just a nice to have and will go out whenever we push a new release.

### DIFF
--- a/src/operators/suffix.ts
+++ b/src/operators/suffix.ts
@@ -52,6 +52,8 @@ export class SuffixOperator implements Operator {
 
   readonly maxProps: number;
 
+  static readonly MaxPropsCeiling = 5000; // set a ceiling on maxProps relative to the maximum cardinality in FullStory
+
   constructor(public options: SuffixOperatorOptions) {
     // NOTE the index is -1 because payloads to FS.event or FS.setUserVars are the last in the list of args
     const { index = -1, maxDepth = 10, maxProps = SuffixOperator.DefaultMaxProps } = options;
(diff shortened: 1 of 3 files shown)


src/operators/suffix.ts
```diff
@@ -52,13 +52,20 @@ export class SuffixOperator implements Operator {
...
     // NOTE the index is -1 because payloads to FS.event or FS.setUserVars are the last in the list of args
     const { index = -1, maxDepth = 10, maxProps = SuffixOperator.DefaultMaxProps } = options;
 
     this.index = index;
     this.maxDepth = maxDepth;
     this.maxProps = maxProps;
+
+    if (this.maxProps >= SuffixOperator.MaxPropsCeiling) {
+      // run the validator, which will trigger and error to be thrown
+      this.validate();
```
sabrina-li, line 67: curious the reason for not "always validate" vs only validate when `this.maxProps >= SuffixOperator.MaxPropsCeiling` 
also seems like if `maxProps` takes the default value it gets validated? Did I understand that right?

van-fs: DLO has a config option to skip validation.  Originally, the idea was to boost performance if rules had been implemented, validated, and been stable.  It's possible that `validate()` never gets called in that scenario.  So what this does is check if the user input is above the ceiling, if so, run the `validate()` function.  This is where the same check is made but causes an error to be thrown.  

> also seems like if maxProps takes the default value it gets validated?
`maxProps` defaults to `DefaultMaxProps`, which is `100`.  So it shouldn't call `validate()` when initializing in the constructor alone. LMK if I'm misunderstanding.

sabrina-li: Ah I see. Thanks for the info!
> maxProps defaults to DefaultMaxProps, which is 100

lol I totally read that wrong.

LGTM!

sabrina-li (approved):

PR #145, opened 2021-08-18
Add configurability over rule registration

van-fs:
This PR adds configuration for rule registration related to timing issues.  There are two approaches using a new `waitUntil` option when declaring rules:
- set `waitUntil` to a number; this simply delays registration N milliseconds
- set `waitUntil` to a `function`; the function is a predicate test on the data layer itself

The retry logic was improved.  It uses exponential backoff with a configurable `maxRetries` limit.  I also made the retry multiplier 250ms to make the intervals more "understandable".  For example, the 8 retry intervals has the following delays (in milliseconds): 250, 500, 1000, 2000, 4000, 8000, 16000, 32000, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Each rule provides a set of options for configuration.  Options with an asterisk
 | `debug` | `false` | Set to true if the rule should print debug for each operator transformation. |
 | `description` | `undefined` | Text description of the rule. |
 | `id` | `undefined` | Unique identifier for the rule. |
-| `maxRetry` | `5` | The maximum number of attempts to search for a missing data layer or test the `waitUntil` predicate. |
+| `maxRetry` | `5` | The maximum number of attempts to search for an `undefined` data layer or test the `waitUntil` predicate. |
 | `monitor` | `true` | Set to true to monitor property changes or function calls |
 | `operators` | `[]` | List of operators that transform data before a destination. |
 | `readOnLoad` | `false` | Rule-specific override for `window[‘_dlo_readOnLoad’]`. |
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Sensitive, private, and confidential information should never be added to a data
 
 DLO is a JavaScript asset that is included on a web page.  FullStory hosts versions of DLO on our CDN.  Versioned releases have the naming convention `<version>.js`, and the most recent version is named `latest.js`:
 
-- https://edge.fullstory.com/datalayer/v1/v1.8.0.js
+- https://edge.fullstory.com/datalayer/v1/v1.10.0.js
 - https://edge.fullstory.com/datalayer/v1/latest.js
 
 If you would like the most up to date version of DLO on your site always, use `latest.js`.  If you'd rather use stable releases and perform manual upgrades, use `<version>.js`.
@@ -147,10 +147,12 @@ Each rule provides a set of options for configuration.  Options with an asterisk
 | `debug` | `false` | Set to true if the rule should print debug for each operator transformation. |
 | `description` | `undefined` | Text description of the rule. |
 | `id` | `undefined` | Unique identifier for the rule. |
+| `maxRetry` | `5` | The maximum number of attempts to search for a missing data layer or test the `waitUntil` predicate. |
 | `monitor` | `true` | Set to true to monitor property changes or function calls |
 | `operators` | `[]` | List of operators that transform data before a destination. |
 | `readOnLoad` | `false` | Rule-specific override for `window[‘_dlo_readOnLoad’]`. |
 | `url` | `undefined` | Specifies a regular expression that enables the rule when the page URL matches. |
+| `waitUntil` | `undefined` | Waits a desired number of milliseconds or predicate function's truthy return type before registering the rule. |
 
 > **Tip:** Use `url` to limit when data is read from a data layer and enhance performance.
 
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -349,7 +349,7 @@ export class DataLayerObserver {
       url,
       monitor = true,
       waitUntil,
-      maxRetry = 5, // with a 250ms wait time means this times out after ~8000ms
+      maxRetry = 5, // with a 250ms wait time means this times out after ~8000ms (see backoff logic in this function)
     } = rule;
 
     // rule properties override global ones
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -329,13 +329,42 @@ export class DataLayerObserver {
   }
 
   /**
+   * Schedules rule registration on a specified delay. The delay will be contingent on the number of attempts and the wait time.
+   * Each subsequent attempt at registration will use exponential backoff to lengthen the delay time between individual attempts.
+   * @param rule DataLayerRule to be registered
+   * @param attempt Numeric attempt counter, which starts at `1`
+   * @param wait Time in milliseconds to use as backoff multiplier
+   */
+  private registerRuleWithDelay(rule: DataLayerRule, attempt = 1, wait = 250) {
+    const { id, source } = rule;
+
+    // with a 250ms wait time means this times out after ~8000ms (see backoff logic in this function)
+    const { maxRetry = 5 } = rule;
+
+    // exponentially back-off with a slight offset to prevent tight grouping of re-registration
+    // NOTE, use `attempt - 1` because the first attempt should be equal to `Math.pow(2, 0)`
+    const delay = (2 ** (attempt - 1) * wait) + Math.random();
+
+    if (attempt > maxRetry) {
+      // give up trying to find the data layer
+      Logger.getInstance().warn(LogMessageType.RuleRegistrationError, {
+        rule: id, source, reason: 'max retries attempted',
+      });
+    } else {
+      setTimeout(() => {
+        this.registerRule(rule, attempt + 1);
+      }, delay);
+    }
+  }
+
+  /**
    * Registers a data layer rule. Assuming the rule's `url` is valid, this results in the rule
    * being parsed, adding a DataHandler with any Operators, registering a source and
    * destination, and monitoring for changes or function calls.
    * @param rule to parse and process
    * @throws error if the rule has missing data or an error occurs during processing
    */
-  registerRule(rule: DataLayerRule, attempt = 1, wait = 250) {
+  registerRule(rule: DataLayerRule, attempt = 1) {
     const { readOnLoad: globalReadOnLoad } = this.config;
 
     const {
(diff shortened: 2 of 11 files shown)


src/observer.ts
```diff
@@ -361,22 +367,49 @@ export class DataLayerObserver {
...
+              this.registerRule({
+                ...rule,
+                waitUntil: 0,
+              }, attempt);
+            }, waitUntil > -1 ? waitUntil : 0); // error check a negative value and schedule immediately if used
+            break;
+          case 'function':
+            if (!waitUntil(target.value)) {
+              // NOTE this is handled and rescheduled by the `catch` block
+              throw Error('Rule predicate test is falsy');
```
patrick-fs, line 389: wouldn't `waitUntil` return a false value if the predicate invocation returned false? I.e. if the test is the existence of an object, and that object doesn't exist yet (b/c it hasn't loaded) then you'd expect `waitUntil` to return false - but this  logic treats a false value as an exception. I think I'm getting tripped up b/c this logic uses error handling to handle expected use-cases. It might be cleaner/more interpretable if the `setTimeout` functions were handled in the main `try` block. WDYT?

README.md
```diff
@@ -147,10 +147,12 @@ Each rule provides a set of options for configuration.  Options with an asterisk
 | `debug` | `false` | Set to true if the rule should print debug for each operator transformation. |
 | `description` | `undefined` | Text description of the rule. |
 | `id` | `undefined` | Unique identifier for the rule. |
+| `maxRetry` | `5` | The maximum number of attempts to search for an `undefined` data layer or test the `waitUntil` predicate. |
 | `monitor` | `true` | Set to true to monitor property changes or function calls |
 | `operators` | `[]` | List of operators that transform data before a destination. |
 | `readOnLoad` | `false` | Rule-specific override for `window[‘_dlo_readOnLoad’]`. |
 | `url` | `undefined` | Specifies a regular expression that enables the rule when the page URL matches. |
+| `waitUntil` | `undefined` | Waits a desired number of milliseconds or predicate function's truthy return type before registering the rule. |
```
patrick-fs, line 155: I wonder if waiting a set amount of time is a necessary feature. If a customer were to use this, they'd have to anticipate timing of function executions in the browser, which can be impacted by any number of things, in which case the `waitUntil` millisecond value will be inaccurate and the rule won't register. What if you just made this a predicate function?

src/observer.ts
```diff
@@ -361,22 +367,49 @@ export class DataLayerObserver {
...
+            setTimeout(() => {
+              // re-register the rule so that it schedules immediately upon next invocation of `registerRule`
+              this.registerRule({
+                ...rule,
+                waitUntil: 0,
+              }, attempt);
+            }, waitUntil > -1 ? waitUntil : 0); // error check a negative value and schedule immediately if used
+            break;
+          case 'function':
+            if (!waitUntil(target.value)) {
```
patrick-fs, line 387: Thinking out loud: but since you're already operating on the target value, and this value is a thing that shouldn't be `undefined` - you might consider just implementing `waitUntil` as an internal operation that runs on every rule registration - it basically tests `target.value !== undefined`

src/observer.ts
van-fs: I agree.  Let me revisit with this context.

README.md
van-fs: One use case where the time-based delay helps is building Page Vars for A/B tests.  I'm seeing that we often have lots of these and giving it a 2-3 second delay let's the data layer populate.  The difficulty with a predicate approach is that checking if a single test exists in the data layer results in firing N-1 events.  Not knowing how many events could be in the data layer makes if difficult to write the predicate test beforehand.

The other difficulty - and this one I'm keenly interested in because it applies to a current project - is that we can't store predicate functions due to JSON unmarshalling. So we'll lose serialization if we just have it.  We could combat that problem with some pre-defined checks (e.g. `hasProperties`).  Or a safe form of [eval](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#never_use_eval!).  If the serialization were a non-issue, I'd definitely go with just the predicate function itself.  Maybe that's the path to pursue.

src/observer.ts
```diff
@@ -324,14 +328,43 @@ export class DataLayerObserver {
...
   }
 
+  /**
+   * Schedules rule registration on a specified delay. The delay will be contingent on the number of attempts and the wait time.
+   * Each subsequent attempt at registration will use exponential backoff to lengthen the delay time between individual attempts.
+   * @param rule DataLayerRule to be registered
+   * @param attempt Numeric attempt counter, which starts at `1`
+   * @param wait Time in milliseconds to use as backoff multiplier
+   */
+  private registerRuleWithDelay(rule: DataLayerRule, attempt = 1, wait = 250) {
```
van-fs, line 338: Patrick, refactored out the registration logic.  Feels like this makes things a bit cleaner.

src/observer.ts
```diff
@@ -362,21 +398,26 @@ export class DataLayerObserver {
 
     try {
       const target = DataLayerTarget.find(source);
```
van-fs, line 400: The `DataLayerTarget.find` no longer throws.  It's still in a try catch for unexpected issues, but it no longer throws for situations where the data layer is `undefined`.

src/observer.ts
```diff
@@ -362,21 +398,26 @@ export class DataLayerObserver {
...
-      // schedule subsequent attempts at (attempt * wait) later
-      setTimeout(() => {
-        try {
-          const target = DataLayerTarget.find(source);
-          this.registerTarget(target, operators, destination, readOnLoad, monitor, debug, debounce);
-        } catch (err) {
-          if (attempt > 3) {
-            Logger.getInstance().error(LogMessageType.RuleRegistrationError, { rule: id, source, reason: err.message });
+
+      switch (typeof waitUntil) {
```
van-fs, line 402: Internally the `registerRule` defaults to a `waitUntil` function that checks `undefined` as well as the data layer is of the supported types.

src/observer.ts
```diff
@@ -362,21 +398,26 @@ export class DataLayerObserver {
...
+        case 'number':
+          // NOTE this delay is scheduled *after* the data layer is found to be defined on the page (not after page load)
+          setTimeout(() => {
```
van-fs, line 405: Rather than re-use `registerWithDelay` if the delay is a specific number of milliseconds.  Trying to re-use `registerWithDelay` for this use case creates another call to `registerRule`, which is circuitous and likely unexpected.

van-fs:
@patrick-fs Re-worked the code a bit.  There's one use case I want to tee up for discussion.  If you create a selector `s[^(eVar)]` for example, you're desiring to monitor just eVars in the `s` object.  To me, that seems like we should treat an `s` object with no eVar properties as being "not ready" and begin `waitUntil`.  What I'm pondering is whether we should just make a generic case that a data layer with no properties (i.e. `0` properties) is likely "not ready" or we should be more robust and say there missing properties _that you've included in your selector_ constitutes a "not ready" data layer.  I think our selector feature already supports the latter FWIW.

van-fs:
I think we do need to consider the selector.  If I use `s[^(eVar)]` as the source, this is how we add monitors on a finite list of properties.  In this example, it's all `eVar` prefixed properties (e.g. eVar1, eVar25, etc).  Historically, we at one time just used `s` as the source, but then we create monitors on all sorts of things we don't actually want to monitor (e.g. functions).  I've added an initial solution to the PR.  Will contemplate side effects to consider.

van-fs:
@patrick-fs I haven't had any additional considerations come to mind since making the selector-based change.  The solution is flexible in that any property in the selected object satisfies; it's not all or nothing.  For example, the out-of-the-box Tealium selector is `utag.data[^(brand_,browse_,cart_,category_,customer_,language_,order_,page_,product_,search_,site_,tealium_event)]` but only one of those properties needs to exist before we consider the data layer ready.  The PR feels ready for another review.

src/observer.ts
patrick-fs: Check out this PR: https://github.com/fullstorydev/fullstory-data-layer-observer/pull/146

I've removed that circular dependancy.

src/observer.ts
```diff
@@ -328,14 +328,43 @@ export class DataLayerObserver {
...
+
   /**
    * Registers a data layer rule. Assuming the rule's `url` is valid, this results in the rule
    * being parsed, adding a DataHandler with any Operators, registering a source and
    * destination, and monitoring for changes or function calls.
    * @param rule to parse and process
    * @throws error if the rule has missing data or an error occurs during processing
    */
-  registerRule(rule: DataLayerRule, attempt = 1, wait = 250) {
+  registerRule(rule: DataLayerRule, attempt = 1) {
```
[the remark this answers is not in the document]
patrick-fs: Update: this parameter is removed in this PR: https://github.com/fullstorydev/fullstory-data-layer-observer/pull/146

README.md
patrick-fs: > The difficulty with a predicate approach is that checking if a single test exists in the data layer results in firing N-1 events.

Do you mean that checking to see if a single A/B test exists in the data layer results in firing N-1 events? 

Do you have an example handy? Only asking because I'm trying to build my brain on this and I'm having trouble following.

src/observer.ts
patrick-fs: Thank you. I've iterated on this. WDYT? https://github.com/fullstorydev/fullstory-data-layer-observer/pull/146

README.md
van-fs: Sent a Google Doc with a more concrete example of where we've had a need to reduce N records in a data layer to a single object payload.

patrick-fs: So what I gather is that all properties of the `source` will be operated on (question: does it matter if `readOnLoad` is true or false?)

I'm thinking this behavior is separate from `waitUntil` behavior. But I guess what you're suggesting is that you can apply DataLayer rules with `readOnLoad = false` and a `waitUntil` millisecond value to then operate on all of the experiment data which would have been loaded into the data layer. Do I have that right @van-fs?

patrick-fs (approved):
LGTM - I have a question for you in the thread. Please take a look.

README.md
van-fs: > So what I gather is that all properties of the source will be operated on ...
Correct.  And to answer your question,`readOnLoad` could be either true of false.  It influences behavior after the `waitUntil` condition has been achieved.

Conceptually, `waitUntil` should support some delayed registration/activation of a rule.  In the ideal case, we have the predicate to be an absolute indicator.  For everything else, we can fall back to a naive time-based delay.  I think having an option of last resort may be beneficial.

patrick-fs: > Correct. And to answer your question,readOnLoad could be either true of false. It influences behavior after the waitUntil condition has been achieved.

Ok, I think I i wasn't totally grokking `readOnLoad`. When `true` operators do their thing and destination events are called as soon as the data layer rules are applied. When `false` the operators do their thing and the destination events are called when a mutation happens for the selected property (selector defined in the rule, or inferred from the source).

Do I finally get it? :)

van-fs: Yes, exactly!  Over time the startup logic has become more sophisticated.  What we're doing now is delaying "meaningful" registration of the rule until the data layer is ready.  Originally, the notion of an `onload` event made sense:  DLO boots up, registers rules ... it should read then.  Now we see all sorts of timing issues, missing data layers, data layers that are being constructed, etc.  Maybe we could alias `readOnLoad` to be a `readOnReady` or something.  Not needed for this PR, but if the `onload` event seems opaque (and it sounds like it might be), probably need to revisit.

PR #146, opened 2021-08-27
replace registerRuleWithDelay with a more generic sleep function

patrick-fs:
This PR removes the need to carry the `attempt` parameter around with `registerRule` and the circular dependency between `registerRule` and `registerRuleWithDelay`

@van-fs please check it out and let me know what you think.


van-fs (approved):
LGTM

PR #147, opened 2021-09-02
Adjust suffixing undefined values

van-fs:
I'm finding the suffix operator's handling of undefined needs tuning:
- The large object fix is counting all properties in an object. It appears Adobe for example may have a long list of undefined values (e.g. eVars).  So we're including them in the total count; however, they'll never be sent to the destination because suffix prunes undefined values.
- I added some quick guards for conditions where we just shouldn't process at all. For example, if a property is undefined, just return null so we don't perform needless work.  Same for someone giving us a null/undefined object.

@patrick-fs WDYT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 1.10.1
+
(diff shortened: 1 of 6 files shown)


GGonryun (approved):
LGTM!

PR #148, opened 2021-10-13
Add CEDDL out-of-the-box ruleset; test against this ruleset

ndonze-fs:
Previously, CEDDL tests were against an example ruleset that didn't match the out-of-the-box ruleset provided for organizations which enabled the Data Layer Observer integration and toggled on the Customer Experience Digital Data Layer: 1.0 option. Now, we test against the default ruleset. This required removing and updating a number of tests since some expectations were no longer valid.

There are no behavior changes in this PR, so the package version hasn't been bumped.

/cc @patrick-fs @van-fs

### DIFF
--- a/test/rules-ceddl-fullstory.spec.ts
+++ b/test/rules-ceddl-fullstory.spec.ts
@@ -1,62 +1,24 @@
 import 'mocha';
 
-import { rules } from '../examples/rules/ceddl-fullstory.json';
+import '../rulesets/ceddl.js';
 
 import { basicDigitalData } from './mocks/CEDDL';
 import {
-  expectEqual, expectRule, expectFS, expectMatch, expectType, expectUndefined, ExpectObserver, setupGlobals,
+  expectEqual, expectRule, expectFS, expectMatch, expectUndefined, ExpectObserver, setupGlobals,
   expectGlobal,
 } from './utils/mocha';
+
+const ceddlRulesKey = '_dlo_rules_ceddl';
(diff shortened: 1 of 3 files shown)


test/rules-ceddl-fullstory.spec.ts
```diff
@@ -88,29 +50,6 @@ describe('CEDDL to FullStory rules', () => {
     expectEqual(payload.promotion, 'LaborDay2020');
   });
 
-  it('it should convert strings to reals and send CEDDL cart properties to FS.event', () => {
```
van-fs, line 91: @ndonze-fs, just confirming, the remove of this test case is because it's more or less the same as `it should send CEDDL cart to FS.event`?  If so, I'm good with that.

ndonze-fs:
> LGTM! Just for my understanding: seems like we now have both `rulesets` and `examples`. Only testing against `rulesets` seems fine since that's the ootb rules. I wonder if there's still value in the examples and/or running tests against the example rules. (like tests for the `setUserVars` stuff are no longer included in the ootb rules - which makes sense, we don't want none of thos PPI data 😁)

@sabrina-li I think this is a great callout. If we're providing examples, it feels like we should be verifying those examples behave as expected. As @van-fs mentioned we could also consider deprecating examples if they're no longer valuable. Does it make sense for me to open another another work item to capture/test OOTB Adobe and Tealium rulesets and remove the content under `examples/rules`? Would it make sense to go further and remove all root `examples` folder content (including `.html` examples under `basic-embed`, `ceddl-fs`, and `google-tags-fs`?

PR #151, opened 2021-11-12
Replace String.prototype.startsWith usage with internal implementation

ndonze-fs:
IE11 doesn't support the ES2015 `String.prototype.startsWith` function.

I opted for an internal `startsWith` implementation for a couple reasons:
- I think `startsWith` reads clearer than introducing `indexOf('target', 'searchString') === 0` or `indexOf('target', 'searchString') !== 0` at the current call sites
- I don't think it makes sense for DLO to poly-fill by adding onto `String.prototype` which could impact other JavaScript on the page

Suggestions for alternative approaches definitely welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 1.10.2
+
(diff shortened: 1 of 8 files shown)


sabrina-li (approved):
LGTM

van-fs (approved):
LGTM!

PR #153, opened 2021-11-19
Don't handle empty data in operators or send to configured rule destination

ndonze-fs:
Within `handleData` seemed to be the most optimal place to change this behavior for a couple reasons:
- If we added empty data checking in `handleEvent`, we also would've needed to add empty data checking in `fireEvent` for `readOnLoad: true` scenarios (and we would've needed to test both `fireEvent` and `handleEvent`)
- It's possible data received in `fireEvent` or `handleEvent` could start as non-empty but eventually become empty through operator data processing -- meaning we'd still have a seam to submit empty data to the configured rule destination (e.g. `FS.event`).

Definitely open to feedback about other ways to implement this.

I opted to unit test this pretty heavily and add more of an integration test within our Adobe ruleset tests since empty `eVar` objects sparked interest in this change.

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -104,9 +104,10 @@ export default class DataHandler {
       const { options: { name } } = this.operators[i];
 
       try {
-        // if the data is null, it is a signal to stop processing
+        // if the data is null or empty, it is a signal to stop processing
         // this can happen if an upstream handler needed to prevent a downstream operator
-        if (handledData === null) {
+        // and we don't want to submit empty events to the configured destination
+        if (handledData === null || DataHandler.isEmptyData(handledData)) {
           this.runDebugger(`[${i}] ${name} halted`, handledData, '  ');
           return null;
         }
@@ -145,6 +146,28 @@ export default class DataHandler {
     this.runDebugger(`${path} handleData exit`, handledData);
 
     return handledData;
+  }
+
+  /**
+   * Returns false if there are any items in the data array which have first level child properties
+   * with defined values. Otherwise returns true. Returns true for empty data arrays.
+   * @param data the data array to be handled by operators
+   */
+  private static isEmptyData(data: any[]): boolean {
+    for (let i = 0; i < data.length; i += 1) {
+      const datum = data[i];
+
+      if (typeof datum !== 'object') {
+        return false;
+      }
+
+      // We don't use Object.values because it isn't supported by IE11
+      if (Object.keys(datum).some((key) => datum[key] !== undefined)) {
+        return false;
+      }
+    }
+
+    return true;
   }
 
   /**
--- a/test/handler.spec.ts
+++ b/test/handler.spec.ts
@@ -242,6 +242,8 @@ describe('DataHandler unit tests', () => {
 
       expect(seen.length).to.eq(1);
       expect(seen[0].definedChild).to.eq(val);
+      expect(Object.keys(seen[0])).to.include('undefinedChild');
+      expect(seen[0].undefinedChild).to.be.undefined;
     });
   });
 
--- a/test/handler.spec.ts
+++ b/test/handler.spec.ts
@@ -63,14 +63,18 @@ class GetterOperator implements Operator {
   }
 }
 
-class NullOperator implements Operator {
+class StaticReturnOperator implements Operator {
   options: OperatorOptions = {
-    name: 'null',
+    name: 'static-return',
   };
+
+  constructor(private returnValue: any[] | null) {
+    // sets this.returnValue
+  }
 
   // eslint-disable-next-line class-methods-use-this
   handleData(): any[] | null {
-    return null;
+    return this.returnValue;
   }
 
   validate() {
@@ -122,7 +126,7 @@ describe('DataHandler unit tests', () => {
   it('data layer event data should pass to the first operator', () => {
     const handler = new DataHandler(DataLayerTarget.find('digitalData.page.pageInfo')!);
 
-    const seen: any = [];
+    const seen: any[] = [];
 
     const echo = new EchoOperator(seen);
     handler.push(echo);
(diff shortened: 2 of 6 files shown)


src/handler.ts
```diff
@@ -147,6 +148,28 @@ export default class DataHandler {
...
+  private static isEmptyData(data: any[]): boolean {
+    for (let i = 0; i < data.length; i += 1) {
+      const datum = data[i];
+
+      if (typeof datum !== 'object') {
+        return false;
+      }
+
+      // We don't use Object.values because it isn't supported by IE11
+      if (Object.keys(datum).some((key) => datum[key] !== undefined)) {
```
sabrina-li, line 165: ~do we need to include `!== null` or use `!=`? unsure if setting something to `null` is possibly via data layer~
Seems like we are specifically expecting `null` to be a valid value, just curious why would we want `null`? Is it because `null` should always intentionally set vs. `undefined` isn't?

src/handler.ts
```diff
@@ -104,9 +104,10 @@ export default class DataHandler {
       const { options: { name } } = this.operators[i];
 
       try {
-        // if the data is null, it is a signal to stop processing
+        // if the data is null or empty, it is a signal to stop processing
         // this can happen if an upstream handler needed to prevent a downstream operator
-        if (handledData === null) {
+        // and we don't want to submit empty events to the configured destination
+        if (handledData === null || DataHandler.isEmptyData(handledData)) {
```
sabrina-li, line 110: Why isn't this check needed for line 120? I guess I'm not sure if any (specifically the last in the pipeline) `operator.handleData` could ever return empty not null data?

test/handler.spec.ts
```diff
@@ -194,24 +198,76 @@ describe('DataHandler unit tests', () => {
...
+
+      const seen: any[] = [];
+
+      const echo = new EchoOperator(seen);
+
+      handler.push(echo);
+      handler.fireEvent();
+
+      expect(seen.length).to.eq(1);
+      expect(seen[0].definedChild).to.eq(val);
```
sabrina-li, line 244: should be expecting `undefinedChild` to be dropped? Maybe good to add for clarity?

sabrina-li (commented):
🎉 mostly LGTM. I left some questions, but I'm not as familiar with this code as you do, so sorry if they don't make much sense.

test/handler.spec.ts
ndonze-fs: `undefinedChild` isn't dropped here. It will have a value of `undefined`, but it still still exist in the keys on `seen[0]`. I could be explicit about this behavior by adding:

```
expect(seen[0].undefinedChild).to.eq(undefined);
```

WDYT?

src/handler.ts
ndonze-fs: On line 120 `handledData` is assumed to be an array, and each item in the array is passed back into `handleData` as the `data` arg. This would eventually hit the check on line 110 again as long as there are additional operators configured after `FanOutOperator`. If there aren't additional operators configured, I don't think we'd need to worry about `null` or `empty` data being sent to `FS.event` since data is sent to the configured destination through a `FunctionOperator` at the end of the operator chain.

src/handler.ts
ndonze-fs: This is a good question, and I'm open to suggestions for changes here. I approached this work by trying to make the `readOnLoad` behavior match the behavior for dynamic changes. Dynamic changes treated `null` as an empty object and more of an explicit set value rather than unset, so any `null` values would ultimately get suffixed with `_obj` and sent to `FS.event` as empty objects `{}`.

Maybe @van-fs has context for the original dynamic change behavior and whether this still makes sense.

src/handler.ts
sabrina-li: Oh okay, I missed the fact that the `FunctionOperator` should always be the last in the the operator chain(`this.operators`), good to know! Thx!

test/handler.spec.ts
sabrina-li: I feel like it could be good to be explicit, to document the behavior, as well as preventing us from somehow unintentionally change this behavior in the future. Open to different opinions tho if anyone feels differently

src/handler.ts
ndonze-fs: This is tough to know without digging into the `DataLayerObserver` initialization code. I just came across this recently: https://github.com/fullstorydev/fullstory-data-layer-observer/blob/11c4623525585fdb799d0f753beafc362b6f2550/src/observer.ts#L207

sabrina-li (commented):
Maybe add something to the CHANGELOG?

ndonze-fs:
> Maybe add something to the CHANGELOG?

@sabrina-li  Good catch! Totally missed this.

src/handler.ts
van-fs: For  `handleData`, `null` is used to signal that the operator pipeline should stop processing (and not forward the data to the next operator).  An example of this is if we were to query and receive no result, which should stop processing.  Errors could also halt the pipeline by being caught and `null` returned by the operator.  So `null` has a certain semantics to continue or break.

I think that's why you're seeing `null` as a case for handling data but not the return value for readOnLoad is that when we read the data layer, we're always expecting an object (ideally a non-empty object).  It's worth confirming, but IIRC we have checks on when we monitor and read.  If the target (e.g. digitalData.something) is falsey, we should be backing off or not sending it to `handleEvent`.

test/handler.spec.ts
ndonze-fs: I added a couple assertions to be explicit about expected behavior (key is retained, but value is `undefined`)

ndonze-fs:
@sabrina-li Should be good for another review (and @van-fs if you're intending to review)

PR #154, opened 2021-11-30
Update NPM packages to patch vulnerabilities in dev dependencies

ndonze-fs:


package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@fullstory/data-layer-observer",
-      "version": "1.10.2",
+      "version": "1.11.0",
```
ndonze-fs, line 9: I ran `npm version` to update to 1.11.0, but I suspect this difference was because I was still running an older version of NPM when I ran `npm version` -- and the lock file had already been upgraded to version 2.

patrick-fs: Oh, I see. So `1.10.2` was a stale version and should have been `1.11.0` at this point?

patrick-fs (approved):
LGTM - I have one question for you.

package-lock.json
ndonze-fs: That's my thinking. The package itself was already versioned to `1.11.0` at the top-most level in `package.json` and `package-lock.json`. This change occurred automatically after updating the dev dependency versions and running `npm i`.

PR #155, opened 2021-12-01
Prevent triggering rules with property changes that don't match selector

ndonze-fs:
When applying selector patterns (e.g. prefix / head match) to rule sources it was previously possible for a rule to be triggered by changes to a property on the underlying source even if the property that changed didn't match the rule's selector pattern.

Example: Rule source `s[^(eVar1)]` should only care about changes to `s.eVar1`, `s.eVar10`, `s.eVar100`, etc. However, if another rule were monitoring `s[^(eVar2)]` and the value of `s.eVar2` changed the rule monitoring `s[^(eVar1)]` would've been triggered since both rules monitored the same source `s`. This change partitions eventing by rule source such that rules with different sources are triggered independently by the properties they're monitoring.

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -33,21 +33,23 @@ export class PropertyDetail implements DataLayerDetail {
 /**
  * Defines CustomEvent types. Types will be prefixed with a DLO namespace.
  * See https://developer.mozilla.org/en-US/docs/Web/API/Event/type
+ * @param source source from the rule monitoring the data layer
  * @param path that identifies the data layer object that created the event
  */
-export function createEventType(path: string) {
-  return `datalayerobserver/${path}`;
+export function createEventType(source: string, path: string) {
+  return `datalayerobserver/${source}/${path}`;
 }
 
 /**
  * Builds a CustomEvent used to broadcast changes.
+ * @param source source from the rule monitoring the data layer
  * @param target that triggered the event (see https://developer.mozilla.org/en-US/docs/Web/API/Event/target)
  * @param property that triggered the event
  * @param value that was emitted by the target
  * @param path to the target
  */
-export function createEvent(target: any, property: string, value: any, path: string): CustomEvent<DataLayerDetail> {
-  return new CustomEvent<DataLayerDetail>(createEventType(path), {
+export function createEvent(source: string, target: any, property: string, value: any, path: string): CustomEvent<DataLayerDetail> {
+  return new CustomEvent<DataLayerDetail>(createEventType(source, path), {
     detail: typeof target[property] === 'function' ? new FunctionDetail(path, property, value)
       : new PropertyDetail(path, property, value),
   });
--- a/src/event.ts
+++ b/src/event.ts
@@ -33,7 +33,7 @@ export class PropertyDetail implements DataLayerDetail {
 /**
  * Defines CustomEvent types. Types will be prefixed with a DLO namespace.
  * See https://developer.mozilla.org/en-US/docs/Web/API/Event/type
- * @param source source from the rule monitoring the data layer
+ * @param source from the rule monitoring the data layer
  * @param path that identifies the data layer object that created the event
  */
 export function createEventType(source: string, path: string) {
@@ -42,7 +42,7 @@ export function createEventType(source: string, path: string) {
 
 /**
  * Builds a CustomEvent used to broadcast changes.
- * @param source source from the rule monitoring the data layer
+ * @param source from the rule monitoring the data layer
  * @param target that triggered the event (see https://developer.mozilla.org/en-US/docs/Web/API/Event/target)
  * @param property that triggered the event
  * @param value that was emitted by the target
--- a/src/event.ts
+++ b/src/event.ts
@@ -48,7 +48,13 @@ export function createEventType(source: string, path: string) {
  * @param value that was emitted by the target
  * @param path to the target
  */
-export function createEvent(source: string, target: any, property: string, value: any, path: string): CustomEvent<DataLayerDetail> {
+export function createEvent(
+  source: string,
+  target: any,
+  property: string,
+  value: any,
+  path: string,
+): CustomEvent<DataLayerDetail> {
   return new CustomEvent<DataLayerDetail>(createEventType(source, path), {
     detail: typeof target[property] === 'function' ? new FunctionDetail(path, property, value)
       : new PropertyDetail(path, property, value),
--- a/test/rules-custom.spec.ts
+++ b/test/rules-custom.spec.ts
@@ -2,7 +2,7 @@ import 'mocha';
 
 import { DataLayerRule } from '../src/observer';
 import {
-  expectEqual, waitForFS, setupGlobals, ExpectObserver, expectGlobal,
+  expectEqual, waitForFS, setupGlobals, ExpectObserver, expectGlobal, expectNoCalls,
 } from './utils/mocha';
 
 describe('Custom rules', () => {
@@ -10,7 +10,9 @@ describe('Custom rules', () => {
     ExpectObserver.getInstance().cleanup();
   });
 
-  it('should only trigger rules for property changes that match rule selector', async () => {
+  it('should only trigger rules for property changes that match rule selector', async function test() {
+    (this as any)!.timeout(3000);
+
     const s = {
       eVar11: 'test',
       eVar22: 'test',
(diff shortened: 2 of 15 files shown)


src/event.ts
```diff
@@ -33,21 +33,29 @@ export class PropertyDetail implements DataLayerDetail {
 /**
  * Defines CustomEvent types. Types will be prefixed with a DLO namespace.
  * See https://developer.mozilla.org/en-US/docs/Web/API/Event/type
+ * @param source from the rule monitoring the data layer
```
sabrina-li, line 36: just making sure I understand the change, `source` is the entire string of source in the rule, and `path` is the object without the "selection"? Sorry I'm still a bit fuzzy on the concepts here.
For example, if the rule has a `source: s[^(eVar1)]`, this `source` would be `s[^(eVar1)]` and path is `s`?

ndonze-fs: You've got that right. `source` is taken directly from the `source` value configured on the rule, whereas `path` is the object being monitored. `source` and `path` can be the same, or they can be different. Some examples:

```
source: nestedData.child, path: nestedData.child
source: s[^(eVar1)], path: s
```

`path` is the object that gets shims defined for its properties and functions to trigger rules on property value changes and function calls.

Previously, each property or function could only be shimmed once, and each shim could only trigger one event path. So, if multiple rules applied to the same `path` object (like `s`), all those rules had to be triggered on property value change or function call even if a selector didn't care about the property that changed or the function that was called.

sabrina-li: Okay I'm still wrapping my head around it, so each path/object will still be shimmed once, but any changes get monitored, it will emit an even per source, per path. And then only the rules that matches both will care about the event and process the data? 
^ That SGTM! (with my limited DLO knowledge 😅) Only thing I wonder is how much more "chatty" would this make dlo's custom events, but since rules are manually added, it's probably not a big deal?

ndonze-fs: > so each path/object will still be shimmed once, but any changes get monitored, it will emit an even per source, per path. And then only the rules that matches both will care about the event and process the data?

That all sounds correct to me. In terms of chattiness, DLO will raise more browser events (`window.dispatch`) when multiple rules are monitoring the same path. Since this only affects cases where multiple rules monitor the same path and rules are explicitly defined (via OOTB rulesets or via custom rules) -- and since browser events are typically high volume (mouse movement, key down, key up, click, etc.) -- I'm expecting the impact to be minimal.

In contrast, this change will reduce `FS.event` chattiness for rules monitoring the same path where the data doesn't match a rule's selector pattern.

test/rules-custom.spec.ts
```diff
@@ -0,0 +1,61 @@
+import 'mocha';
```
ndonze-fs, line 1: I added this spec file as a place where we can test edge cases in a more end-to-end fashion which aren't exposed by out-of-the-box rulesets. Interested if there are other recommendations for where/how to test edge cases like these.

test/utils/mocha.ts
```diff
@@ -283,14 +283,46 @@ export function expectRule(id: string, ruleset?: string): DataLayerRule {
...
+
+/**
+ * Similar to `expectFS`, `waitForFS` is a convenicne method for `expectParams` that
+ * checks the global FullStory object. Unlike `expectFS`, `waitForFS` supports waiting
+ * for a global FullStory object method call within a timeout period.
+ * @param methodName FullStory API function (`methodName` arg for `expectParams`)
+ * @param namespace Global object if FullStory is not `FS`
+ * @param timeout Time in milliseconds after which the Promise is rejected
+ */
+export function waitForFS(methodName: FSMethodName, namespace = 'FS', timeout: number = 1000): Promise<any[]> {
```
ndonze-fs, line 310: I added this convenience function to support testing the edge case this PR intends to resolve. To verify this behavior, I needed to change data layer values and wait for event dispatching to carry through calls to rule destinations. This can take a variable duration to propagate, so I've checked the condition every 50ms here up until the configured timeout. Interested if you all have better ideas for how to approach this.

ndonze-fs:
@sabrina-li @van-fs This should be ready for a proper review. If it'd be helpful to walk through any of this synchronously, please let me know.

test/rules-custom.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    // Wait for event dispatch to result in rule destination calls
+    let [eventName, payload] = await waitForFS('event');
+    expectEqual('eVar1', eventName);
+    expectEqual(11, payload.eVar11);
+
+    expectGlobal('s').eVar22 = 22;
+
+    [eventName, payload] = await waitForFS('event');
+    expectEqual('eVar2', eventName);
+    expectEqual(22, payload.eVar22);
```
sabrina-li, line 59: Hmm is it possible to verify that no event is fired with `eventName == eVar1` when `eVar22` changes, and vice versa? Not sure if that's already covered in this test case, or maybe worth adding here?

sabrina-li (approved):
🎉 Overall LGTM, just left some nits and comments for you to checkout

test/rules-custom.spec.ts
ndonze-fs: This is a good suggestion. I originally avoided doing this because I couldn't think of a better strategy than waiting for some defined period of time and checking that no further calls were made. I think this strategy is reasonable, though, considering it takes ~10-20s to run the entire test suite.

test/utils/mocha.ts
ndonze-fs: I think the challenge we'd run into is there are multiple waits within the test -- wait for first rule destination call, wait for second rule destination call, and wait for a defined period of time before ensuring no further rule destinations were called. I don't think we'd be able to call `done()` until the last condition were true, whereas it feels like async/await better handles the stepped nature of the waits. Let me know if any of that doesn't make sense. I might not be fully understanding how we could thread `done` through the wait conditions in this test.

van-fs:
@ndonze-fs Originally, `Monitor` and its implementation `ShimMonitor` were designed to lack knowledge of the outside world.  It "monitored" changes on a property, and if one occurred, broadcasted the event.  At that time `source` and `path` were pretty much the same.  It wasn't until later that `source` became a selector to support other behavior.  Now that knowledge related to the `source` has been pushed deeper into the stack, does the `window.dispatch` model still apply?  What are your thoughts on simply calling `handleData` on the `DataHandler` reference directly rather than going through the window event listener model?

ndonze-fs:
> @ndonze-fs Originally, `Monitor` and its implementation `ShimMonitor` were designed to lack knowledge of the outside world. It "monitored" changes on a property, and if one occurred, broadcasted the event. At that time `source` and `path` were pretty much the same. It wasn't until later that `source` became a selector to support other behavior. Now that knowledge related to the `source` has been pushed deeper into the stack, does the `window.dispatch` model still apply? What are your thoughts on simply calling `handleData` on the `DataHandler` reference directly rather than going through the window event listener model?

@van-fs I like this suggestion! Do you think this makes sense to explore as part of this scope of work, or do you think it makes sense to tackle monitoring and handler invocation changes in a separate PR (i.e. separate the behavior change from the architectural change)?

Initially, I wondered if dispatching events and handling them via listeners gives us any deferral that might be beneficially breaking up DLO work that could block the UI thread, but everything I'm reading suggests `window.dispatchEvent` and subsequent event handling are synchronous and blocking -- so it doesn't seem like we'd be losing anything by transitioning to direct method invocation on the handler.

van-fs:
> I wondered if dispatching events and handling them via listeners gives us any deferral that might be beneficially breaking up DLO work that could block the UI thread

You took the words right out of my mouth.  After I looked at the monitoring/logging story, I thought that it is nice to have something between initialization and handler invocation.  I wasn't sure if the window event gave us that benefit.  We could also explore an async callback model, but it does sound like that's sizable enough to defer to another PR.  The reality is that that DLO runs pretty fast already (single digit milliseconds often), and we'll be able to prioritize this decision with more log data.

> other than an extensibility point that I'm not aware is being used -- other client side code being able to subscribe to property changes

It's nice that it's there, but I doubt anyone is using it in practice.  If anything, I expect developers would likely simply register the rule/operator and use the framework rather than handle the event.  It was an idea we had early that may make sense to wall off down the road in favor of a "supported" extension model.

ndonze-fs:
> After I looked at the monitoring/logging story, I thought that it is nice to have something between initialization and handler invocation. I wasn't sure if the window event gave us that benefit. We could also explore an async callback model, but it does sound like that's sizable enough to defer to another PR.

@van-fs I agree that it may be better to tackle this under a separate scope of work. Beyond thinking about async data handling, it looks like we're also debouncing within `handleEvent` -- so removing event dispatching will likely require some shifting there.

How do you feel about this work as is? Any changes you'd like to see? Good with merging and releasing what's here?

PR #162, opened 2022-03-03
Allowing testing of empty string for selector

irv-fs:
This pull request allows string values to be quoted with `'` or `"` values.  To be backwards compatible, the quotes are optional.  In the code I simply look for them in string values and strip them off if they are there.  Tests were updated to try with and without quotes, as well as testing for empty string.  Empty string can be tested by doing ?(attribute='') in the selector.

### DIFF
--- a/src/selector.ts
+++ b/src/selector.ts
@@ -393,14 +393,25 @@ class PathElement {
         case 'boolean':
           if (prop[opProp.name] !== (opProp.value.toLowerCase() === 'true')) return undefined;
           break;
-        case 'string':
+        case 'string': {
+          let realString = opProp.value;
+          // look for quoted values, and strip the quotes off if they are there
+          if (opProp.value) {
+            const firstCharacter = realString.charAt(0);
+            const lastCharacter = realString.charAt(realString.length - 1);
+            if ((firstCharacter === '"' && lastCharacter === '"')
+              || (firstCharacter === "'" && lastCharacter === "'")) {
+              realString = realString.substring(1, realString.length - 1);
+            }
+          }
           // eslint-disable-next-line eqeqeq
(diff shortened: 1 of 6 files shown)


src/selector.ts
```diff
@@ -393,14 +393,25 @@ class PathElement {
...
+        case 'string': {
+          let realString = opProp.value;
+          // look for quoted values, and strip the quotes off if they are there
+          if (opProp.value) {
+            const firstCharacter = realString.charAt(0);
+            const lastCharacter = realString.charAt(realString.length - 1);
+            if ((firstCharacter === '"' && lastCharacter === '"')
+              || (firstCharacter === "'" && lastCharacter === "'")) {
+              realString = realString.substring(1, realString.length - 1);
+            }
```
irv-fs, line 405: String the matching `'` or `"` characters off the string and set value into realString.  Then use that everywhere else in the string comparison.  Makes it backwards compatible.

test/selector.spec.ts
```diff
@@ -7,6 +7,7 @@ import { select, validate, Path } from '../src/selector';
 const testData = {
```
irv-fs, line 7: Most of this is just taking existing tests and adding in tests for quoting them.  Also empty string tests added.

ndonze-fs (approved):
This mostly LGTM. The one question I have is if we should be testing that a value with a single `'` or a single `"` behaves as expected. This would be an invalid selector; does it behave the same as other invalid selectors / selector parsing issues?

ndonze-fs:
@irv-fs Once we get this merged I think we'll also want to look at a patch version bump for the library and to document this change in `CHANGELOG`. Let me know if you want any help with that. We could include the version bump and `CHANGELOG` entry in this PR or under a separate PR.

ndonze-fs (approved):
LGTM

PR #163, opened 2022-03-03
Add console telemetry exporter

ndonze-fs:
Adds a console telemetry exporter which will be the default telemetry exporter once telemetry collection is wired up. Builds upon work in #161.

Upcoming work:
- Implement "safe" telemetry wrappers which prevent telemetry collection and reporting from throwing errors
- Wire up telemetry to collect metrics (e.g. initialization duration)
- Enable configuring a custom telemetry provider or telemetry exporter

I haven't bumped the library version as part of these changes since this exporter is unused and will not be included in build bundles.

### DIFF
--- a/test/telemetry.spec.ts
+++ b/test/telemetry.spec.ts
@@ -7,9 +7,6 @@ import { MockClass } from './mocks/mock';
 import Console from './mocks/console';
 import { expectNoCalls, expectParams } from './utils/mocha';
 
-const originalConsole = globalThis.console;
-const mockConsole = new Console();
-
 class MockTelemetryExporter extends MockClass {
   sendSpan(): void {}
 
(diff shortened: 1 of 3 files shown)


ndonze-fs:
@djwhatley-fs This is ready for review. There was another test that was changing global state without resetting it. I've fixed that in this PR since it was preventing new telemetry tests from passing as part of a full test run.

test/telemetry.spec.ts
```diff
@@ -13,75 +14,137 @@ class MockTelemetryExporter extends MockClass {
 }
 
 describe('DefaultTelemetryProvider', () => {
```
ndonze-fs, line 16: @djwhatley-fs I recommend hiding whitespace when reviewing this file. The changes to existing tests reduced nesting.

djwhatley-fs (approved):
Nice, LGTM

PR #166, opened 2022-03-07
Clean up whitespace which fails release sync precommit hooks

ndonze-fs:
I haven't bumped the library version since there aren't any behavior changes. This extraneous whitespace trips up precommit hooks as part of synchronizing for releases.


sabrina-li (approved):
LGTM!

PR #167, opened 2022-03-08
Add wiring to support default and custom telemetry providers and exporters

ndonze-fs:
When this PR merges, telemetry will be initialized but is otherwise unused (nothing is measured or exported). This PR:

- Implements error trapping in the default telemetry provider to prevent future telemetry collection from throwing errors when the default telemetry provider is used. The default telemetry provider is used when no custom telemetry provider is configured. This is the case even when a custom telemetry exporter is configured
- Wires up telemetry initialization to DLO initialization
- Accepts a custom telemetry provider or custom telemetry exporter as window configuration similar to other `window._dlo_*` configuration
- Bumps the DLO library minor version to indicate the presence of a new backward compatible feature


src/embed/init.ts
```diff
@@ -79,7 +91,7 @@ function _dlo_collectRules(): any[] {
   }
 }
 
-function _dlo_initializeFromWindow() {
+export default function _dlo_initializeFromWindow() {
```
ndonze-fs, line 94: I'm not thrilled about exporting this function to facilitate testing, but this seemed like a narrow and low risk way to enable testing of telemetry initialization from window configuration. Note: This change results in a new warning when building DLO:

> (!) If you do not supply "output.name", you may not be able to access the exports of an IIFE bundle.

This export is only used from tests and does not seem to cause any testing issues.

src/utils/telemetry.ts
```diff
@@ -207,3 +225,43 @@ export const consoleTelemetryExporter: TelemetryExporter = {
...
+    }
+
+    return Telemetry.instance;
+  }
+
+  /**
+   * Clears the initialized telemetry provider allowing telemetry to be reinitialized. Used
+   * within telemetry intialization tests to achieve a clean starting state
+   */
+  static reset(): void {
```
ndonze-fs, line 264: I'm not enthusiastic about adding a `reset` method strictly for tests. This is one of the challenges of following the singleton pattern we use for logging. For logging, we allow directly modifying the appender and log levels on the singleton logging instance. Since telemetry configuration should not be changed after initializing other than within tests, I opted for a `reset` method instead of exposing telemetry state for manipulation (e.g. changing the provider and/or exporter after initializing).

test/telemetry.spec.ts
```diff
@@ -148,3 +208,150 @@ describe('ConsoleTelemetryExporter', () => {
...
+      .filter((key) => key.startsWith('_dlo_rules'))
+      .forEach((key) => {
+        win[key] = [];
+      });
+
+    // We dynamically import the initialization function so we can remove rules
+    // beforehand. If we import the initialization function while rules from other
+    // tests are still on the global window, test output noise is generated due to
+    // unexpected data layer and rule state
+    initializeFromWindow = (await import('../src/embed/init')).default;
```
ndonze-fs, line 295: This is unfortunate complexity to absorb in test setup here. We have other tests that set `window._dlo_rules*` values and don't clean up those values on the global window. Rather than updating all those tests (which is probably still a good idea), I took a defensive stance in these tests to limit overall PR surface.

src/utils/telemetry.ts
```diff
@@ -207,3 +225,43 @@ export const consoleTelemetryExporter: TelemetryExporter = {
...
+
+  /**
+   * Gets the singleton {@link TelemetryProvider} configured for the Data Layer Observer instance.
+   * Initializes the singleton {@link TelemetryProvider} on first call
+   *
+   * @param provider The {@link TelemetryProvider} to initialize; only effective on first call
+   * @param exporter The {@link TelemetryExporter} to use with the default Telemetry Provider; only
+   * used if no {@link TelemetryProvider} is given and only effective on first call
+   */
+  static getInstance(provider?: TelemetryProvider, exporter?: TelemetryExporter): TelemetryProvider {
```
patrick-fs, line 244: Just a thought: rather than introduce side effects with these parameters, I wonder if it would be better to introfcut two new static methods: `withExporter` and `withProvider`?

then `getInstance` will take no parameters and return the default provider + exporter if neither of the `with` methods are called.

WDYT?

ndonze-fs: Overall I like this idea. How would you propose setting the default instance initially? I agree it's a little odd that `getInstance` is responsible for returning an instance and initially setting the instance -- but only if it's the first call. On subsequent calls, provider and exporter args are ignored.

Maybe we get away from the idea of silently setting the default instance and make it explicit. And, you can change the default instance if you wish (would be useful in tests). Example:

```
Telemetry.setInstance(Telemetry.withProvider(window._dlo_telemetryProvider));
```
or...
```
Telemetry.setInstance(Telemetry.withExporter(window._dlo_telemetryExporter));
```

We don't allow setting both a provider and an exporter, so it's up to the caller (`init` in our case) to determine which call to make. While the caller burden isn't ideal, it does make the inability to pass both a provider and an exporter more explicit.

If we lay out a new pattern here, it might be valuable to consider extending the new pattern to logging in the future. Example:

``
Logger.setInstance(Logger.withAppender('console'));
``

patrick-fs: I like that! 

`Telemetry.setInstance(Telemetry.with_X_())` reads nicely. Love the idea of extending the patter to other static accessors too.

src/utils/telemetry.ts
patrick-fs: `setInstance` (from our prior thread) may solve this problem?

src/utils/telemetry.ts
```diff
@@ -127,12 +128,16 @@ class DefaultTelemetrySpan implements TelemetrySpan {
    * sendSpan callback with a timespan event
    */
   end() {
-    this.sendSpan({
-      name: this.name,
-      timestamp: new Date().toISOString(),
-      attributes: this.attributes,
-      duration: DefaultTelemetrySpan.getCurrentTime() - this.startTime,
-    });
+    try {
```
patrick-fs, line 131: In your experimental branch, you were wrapping providers and spans into `Safe` equivalents to handle `try/catch`. Just curious: what motivated you to move away from that pattern? 

https://github.com/fullstorydev/fullstory-data-layer-observer/compare/nate/experimental-metrics

test/telemetry.spec.ts
```diff
@@ -1,19 +1,48 @@
-/* eslint-disable class-methods-use-this, @typescript-eslint/no-unused-vars */
+/* eslint-disable class-methods-use-this, max-classes-per-file, no-underscore-dangle, @typescript-eslint/no-unused-vars */
```
patrick-fs, line 1: Would it be a good idea to test error handling too? I'm not sure either way; just wanted to get your thoughts.

src/utils/telemetry.ts
ndonze-fs: Good question, and I should've called out this deviation from the proof of concept.

The motivation behind wrapping originally was to offload the burden of error handling from exporters -- allowing exporters to be slimmer and focus on transforming metrics and sending them to some destination. In the proof of concept, I also wrapped any given custom telemetry provider in the safe telemetry provider. On reflection, this felt like it could be somewhat hostile as providing a custom telemetry provider is a stronger "ejection" from the bundled telemetry mechanisms. In contrast, providing only an exporter is opting into the default telemetry provider as a collector and export invoker.

Here, I opted to put the error trapping in the default telemetry provider and allow/force a custom telemetry provider to provide its own error handling. Definitely open to feedback if the original proof of concept feels like a better approach (or something else).

test/telemetry.spec.ts
ndonze-fs: Did you have some examples in mind? There are some tests here which verify behaviors when `sendSpan` or `sendCount` throw errors.

src/embed/init.ts
patrick-fs: Not a huge deal - but it might be cool if there were a way to suppress that warning or document what's causing it and why it's safe to be ignored.

src/utils/telemetry.ts
patrick-fs: It seems like we could stick with the implementation you have in place here and then move towards a "Safe" wrapper implementation in the future if necessary. It doesn't change the default interface one bit and should be harmless to wrap the instance exporters and providers in "safe" wrappers later. I'll lean towards the simpler architecture, which is what you have here.

patrick-fs (approved):
LGTM, pending an update for the static Telemetry interface (`Telemetry.setInstance(Telemetry.with_X_())`)

Thanks!

test/telemetry.spec.ts
patrick-fs: Ah ok, I just missed those.

src/utils/telemetry.ts
```diff
@@ -207,3 +225,62 @@ export const consoleTelemetryExporter: TelemetryExporter = {
...
+    return new DefaultTelemetryProvider(nullTelemetryExporter);
+  }
+
+  /**
+   * Gets the configured {@link TelemetryProvider} instance. Initializes a default
+   * telemetry provider if a telemetry provider hasn't been set
+   */
+  static getInstance(): TelemetryProvider {
+    if (!Telemetry.instance) {
+      Telemetry.instance = Telemetry.withExporter(nullTelemetryExporter);
```
ndonze-fs, line 282: This still has a side effect, but I wanted to make it safe to call this method and get a default provider rather than throwing.

src/embed/init.ts
ndonze-fs: I moved initialization functions to a separate file where they could be exported for both library initialization and simulated initialization from tests without also being coupled to IIFE. This addressed the rollup build warning.

src/utils/telemetry.ts
```diff
@@ -207,3 +225,62 @@ export const consoleTelemetryExporter: TelemetryExporter = {
...
+    Telemetry.instance = provider;
+  }
+
+  /**
+   * Returns a {@link DefaultTelemetryProvider} which sends telemetry to the given
+   * telemetry exporter
+   *
+   * @param exporter The exporter for sending telemetry
+   */
+  static withExporter(exporter: 'console' | TelemetryExporter | undefined) : TelemetryProvider {
```
patrick-fs, line 266: For a future refactor you might consider defining and `WithExporterParam` type or something like that. Not necessary to update this PR, just remarking.

patrick-fs:
> @patrick-fs This should be ready for another review.

@ndonze-fs updates LGTM!

PR #169, opened 2022-03-10
Fixing logLevel bug where ERROR (0) would get dropped

irv-fs:
Due to the previous technique of || the value 0 would never come through.  This allows that value to come through as well as officially trying to parse string values if they contain numbers

### DIFF
--- a/src/embed/init.impl.ts
+++ b/src/embed/init.impl.ts
@@ -130,6 +130,7 @@ export default function _dlo_initializeFromWindow() {
         break;
       case 'string':
         level = parseInt(win._dlo_logLevel, 10);
+        level = Number.isNaN(level) ? undefined : level;
         break;
       default:
         level = undefined;
--- a/src/embed/init.impl.ts
+++ b/src/embed/init.impl.ts
@@ -120,25 +120,10 @@ export default function _dlo_initializeFromWindow() {
       Logger.getInstance().warn(LogMessageType.ObserverRulesNone);
     }
 
-    // get the log level from win._dlo_logLevel
-    // we allow string or number, and use parseInt if it is a string
-    // otherwise it is set to undefined, which will use the default log level
-    let level;
-    switch (typeof win._dlo_logLevel) {
-      case 'number':
-        level = win._dlo_logLevel;
-        break;
-      case 'string':
-        level = parseInt(win._dlo_logLevel, 10);
-        level = Number.isNaN(level) ? undefined : level;
-        break;
-      default:
-        level = undefined;
-    }
     win._dlo_observer = new DataLayerObserver({
       appender: win._dlo_appender || undefined,
       beforeDestination: win._dlo_beforeDestination || undefined,
-      logLevel: level,
+      logLevel: win._dlo_logLevel,
       previewMode: win._dlo_previewMode === true,
       previewDestination: win._dlo_previewDestination || undefined,
       readOnLoad: win._dlo_readOnLoad === true,
--- a/src/embed/init.impl.ts
+++ b/src/embed/init.impl.ts
@@ -120,10 +120,24 @@ export default function _dlo_initializeFromWindow() {
       Logger.getInstance().warn(LogMessageType.ObserverRulesNone);
     }
 
+    // get the log level from win._dlo_logLevel
+    // we allow string or number, and use parseInt if it is a string
+    // otherwise it is set to undefined, which will use the default log level
+    let level;
+    switch (typeof win._dlo_logLevel) {
+      case 'number':
+        level = win._dlo_logLevel;
+        break;
+      case 'string':
+        level = parseInt(win._dlo_logLevel, 10);
+        break;
+      default:
+        level = undefined;
+    }
     win._dlo_observer = new DataLayerObserver({
       appender: win._dlo_appender || undefined,
       beforeDestination: win._dlo_beforeDestination || undefined,
-      logLevel: win._dlo_logLevel || undefined,
+      logLevel: level,
       previewMode: win._dlo_previewMode === true,
       previewDestination: win._dlo_previewDestination || undefined,
       readOnLoad: win._dlo_readOnLoad === true,
--- a/test/observer.spec.ts
+++ b/test/observer.spec.ts
@@ -12,7 +12,7 @@ import {
   expectParams, expectNoCalls, expectCall, ExpectObserver, expectGlobal, expectEqual, setGlobal,
 } from './utils/mocha';
 import { Operator, OperatorOptions } from '../src/operator';
-import { LogEvent, LogLevel } from '../src/utils/logger';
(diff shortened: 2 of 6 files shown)


src/embed/init.impl.ts
```diff
@@ -120,10 +120,25 @@ export default function _dlo_initializeFromWindow() {
...
 
+    // get the log level from win._dlo_logLevel
+    // we allow string or number, and use parseInt if it is a string
+    // otherwise it is set to undefined, which will use the default log level
+    let level;
+    switch (typeof win._dlo_logLevel) {
+      case 'number':
+        level = win._dlo_logLevel;
+        break;
+      case 'string':
```
van-fs, line 131: JW, was there something we documented that indicated a string was valid?

van-fs:
@irv-fs Can you also bump the version in the README and package as a patch release?

src/embed/init.impl.ts
irv-fs: Yes, I did some testing and a string value that was a number actually worked.  It's in the long slack thread.

irv-fs: From the thread...

It looks like whatever you set it to whether 2 or "foo" or "3" gets preserved.  It all then comes down to

```
if (level <= this.level || level === LogLevel.RECORD) {
      this.appender.log({
        level,
        message,
        context,
      });
    }
```
which does do type promotion if it applies.  If it is a string like foo it never returns true so nothing gets logged.  However, if it is a string with a number in it like "2" that does seem to get promoted.  While 0 doesn't work because of the bug, "0" does work.  :upside_down_face:.  However, I can't get around my problem as I'm using the integration, which I"m guessing most of our customers are using, and that uses a numerical value.

ndonze-fs: I think we can make this simpler _if_ we don't want to make a decision about changing broader log level config behavior and instead want to narrowly focus on fixing this bug.

We could consider passing `win._dlo_logLevel` directly to the `DataLayerObserver` constructor:
```
    win._dlo_observer = new DataLayerObserver({
      ...
      logLevel: win._dlo_logLevel,
      ...
    });
```

Then, in the `DataLayerObserver` constructor, we could explicitly check against `undefined` instead of checking that `logLevel` is truthy (as a bonus, this is compatible with the type being `LogLevel | undefined`):
```
    if (logLevel !== undefined) {
      Logger.getInstance().level = logLevel;
    }
```

Then, we could capture the current behavior with tests. Something like the following in `observer.spec.ts`:
```
  [
    { configLevel: 0, expectedLevel: 0 },
    { configLevel: 1, expectedLevel: 1 },
    { configLevel: 2, expectedLevel: 2 },
    { configLevel: 3, expectedLevel: 3 },
    { configLevel: undefined, expectedLevel: () => Logger.getInstance().level },
  ].forEach((tc) => {
    it(`sets the global log level to ${tc.expectedLevel} given config value ${tc.configLevel}`, () => {
      const observer = ExpectObserver.getInstance().create({
        logLevel: tc.configLevel as number,
        rules: [],
      });

      const expectedLevel = typeof tc.expectedLevel === 'function' ? tc.expectedLevel() : tc.expectedLevel;
      expect(Logger.getInstance().level).to.equal(expectedLevel);

      ExpectObserver.getInstance().cleanup(observer);
    });
  });
```

This would pave the way for us to define expected string or out of bounds behaviors (e.g. `-1` or `4`) in the future without addressing those behavior changes now.

WDYT? /cc @van-fs

irv-fs: That looks fine to me.  It looks like you have the code already written, should we have you just push that, or do you want me to update the code from it?

van-fs: @irv-fs Could you take Nate's suggestion and update the PR?  It sounds like the recommendations are:
1. Keep the init.impl.js shallow and directly pass the expando.
2. Handle the assignment in the DataLayerObserver constructor as-is.

ndonze-fs (approved):
@van-fs: @irv-fs  and I connected offline, and we agreed that I'd push the changes I worked up locally. You may want to review as well since I shouldn't be reviewing my own work. 😀

test/observer.spec.ts
```diff
@@ -413,6 +413,28 @@ describe('DataLayerObserver unit tests', () => {
...
+        logLevel: tc.configLevel,
+        rules: [],
+      });
+
+      expect(Logger.getInstance().level).to.equal(expectedLevel);
+
+      ExpectObserver.getInstance().cleanup(observer);
+    });
+  });
+
```
irv-fs, line 437: In these tests, I would also test string values that are numbers, as those work today, and we wouldn't want to break that.  Your code hasn't broke those, but we should add them to the test.  We should also test that a non numeric string value results in undefined.

ndonze-fs: The interesting part about this is we'll need to set both the config and expected variants to strings unless we do a loose equality comparison in the test -- e.g.:

```
    { configLevel: '3', expectedLevel: '3' },
```

The reason for this is that the string value isn't coerced when assigned; rather, it's coerced when compared to determine if we should write a log entry:

```
    if (level <= this.level || level === LogLevel.RECORD)
```

At this time, a string would be compared with a number, and the string would be coerced to a number.

At least that's the behavior I was seeing locally. Please let me know if I'm missing something or if you're seeing something different.

van-fs: @ndonze-fs Could we keep the tests you have above and consider that as validation that the constructor accepts and sets a value?  That ensures we know the value passed from the window is updated appropriately to the Logger.

Related to coercion, a different test could be added to `logger.spec.ts` that verifies numeric and their string equivalents log appropriately.

Thoughts?

ndonze-fs: @van-fs We could definitely assert this behavior in 2 parts:

- The `DataLayerObserver` constructor does the right thing with the config value provided on the `window` object. As you mentioned, this is captured in the current tests aside from the string variants as @irv-fs pointed out
- Given a string log level value, we log appropriately by level (not captured in tests yet)

Testing for the latter is possible, but may be a little clunky since we'll have to assign a string value to the global log level (casting the type to `number` or `LogLevel` so TypeScript accepts it), call the logger, and then check that methods (e.g. on a mock console instance) were or weren't called based on the log level. I'd be okay moving forward with this if it make sense to you all even though it's not ideal.

Longer term, I'd like to lean into the fact that we don't know what's coming in on the window object and have a `readConfig(window)` or similar which does the heavy lifting of validating config values, converting to appropriate types (e.g. string to number as we'd want for log level), and sets appropriate defaults. I don't think we should pursue that under this scope of work, though.

ndonze-fs: @irv-fs How do you feel about merging this work? Let me know if my comments above didn't address your concern or if you're seeing a different behavior.

irv-fs: I'm good with it!

PR #171, opened 2022-03-21
Instrument significant timings, rule counts, and error counts

ndonze-fs:
This PR uses telemetry to instrument the following:
- Overall DLO initialization time
- Time to collect configured rules from the `window` object
- Time to register all rules
- Time to handle each event (execute operator pipeline and send to configured rule destination) along with the number of operators and type of operators configured for a given event
- Count of client errors segmented by error type
- Count of rules configured for the DLO instance

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,4 +1,5 @@
-import {terser} from 'rollup-plugin-terser';
+import json from '@rollup/plugin-json';
+import { terser } from 'rollup-plugin-terser';
 import typescript from '@rollup/plugin-typescript';
 
 export default {
@@ -6,25 +7,27 @@ export default {
   output: [
     {
       format: 'iife',
-      file: 'build/dlo.js'
+      file: 'build/dlo.js',
     },
     {
       file: 'build/dlo.min.js',
       format: 'cjs',
       name: 'version',
       plugins: [terser({
-        enclose: true
-      })]
-    }
+        enclose: true,
+      })],
+    },
   ],
   plugins: [
+    json(),
     typescript({
-      target: "es5",
-      module: "es2015",
+      target: 'es5',
+      module: 'es2015',
       declaration: false,
       declarationMap: false,
       sourceMap: false,
-      resolveJsonModule: false
-    })
-  ]
+      resolveJsonModule: true,
+      moduleResolution: 'node',
+    }),
+  ],
 };
--- a/src/embed/init.impl.ts
+++ b/src/embed/init.impl.ts
@@ -2,7 +2,9 @@
 import { Logger, LogMessageType } from '../utils/logger';
 import { startsWith } from '../utils/object';
 import { DataLayerObserver } from '../observer';
-import { defaultDloAttributes, Telemetry, telemetryType } from '../utils/telemetry';
+import {
+  defaultDloAttributes, errorAttributes, errorType, Telemetry, telemetryType,
+} from '../utils/telemetry';
 
 /*
 This is where we initialize the DataLayerObserver from this info:
@@ -87,6 +89,7 @@ function _dlo_collectRules(): any[] {
     return results;
   } catch (err) {
     Logger.getInstance().error(LogMessageType.RuleRegistrationError, { reason: `Error: ${err}` });
+    Telemetry.count(telemetryType.clientError, 1, errorAttributes(errorType.ruleRegistrationError));
     return [];
   }
 }
@@ -139,5 +142,6 @@ export default function _dlo_initializeFromWindow() {
     initializationSpan.end();
   } catch (err) {
     Logger.getInstance().error(LogMessageType.ObserverInitializationError, { reason: `Error: ${err}` });
+    Telemetry.count(telemetryType.clientError, 1, errorAttributes(errorType.observerInitializationError));
   }
 }
--- a/src/embed/init.impl.ts
+++ b/src/embed/init.impl.ts
@@ -2,7 +2,7 @@
 import { Logger, LogMessageType } from '../utils/logger';
 import { startsWith } from '../utils/object';
 import { DataLayerObserver } from '../observer';
-import { Telemetry } from '../utils/telemetry';
+import { defaultDloAttributes, Telemetry, telemetryType } from '../utils/telemetry';
 
 /*
 This is where we initialize the DataLayerObserver from this info:
@@ -67,7 +67,7 @@ window['_dlo_rulesFromOpsTeam'] = [
 
 function _dlo_collectRules(): any[] {
   try {
-    const startTime = Date.now();
+    const ruleCollectionSpan = Telemetry.startSpan(telemetryType.ruleCollectionSpan);
     const results: any[] = [];
     Object.getOwnPropertyNames(window).forEach((propName) => {
       if (startsWith(propName, '_dlo_rules') === false) return;
@@ -83,7 +83,7 @@ function _dlo_collectRules(): any[] {
         results.push(rule);
       });
     });
-    Logger.getInstance().record('DLO rule processing time', { numericValue: startTime - Date.now() });
+    ruleCollectionSpan.end();
     return results;
   } catch (err) {
     Logger.getInstance().error(LogMessageType.RuleRegistrationError, { reason: `Error: ${err}` });
@@ -93,7 +93,6 @@ function _dlo_collectRules(): any[] {
 
 export default function _dlo_initializeFromWindow() {
   try {
-    const startTime = Date.now();
     const win = (window as { [key: string]: any });
 
     /*
@@ -104,10 +103,16 @@ export default function _dlo_initializeFromWindow() {
 
     // Logging must be initialized before telemetry since telemetry errors may be logged
     if (win._dlo_telemetryProvider) {
-      Telemetry.setInstance(win._dlo_telemetryProvider);
+      Telemetry.setProvider(win._dlo_telemetryProvider);
     } else {
-      Telemetry.setInstance(Telemetry.withExporter(win._dlo_telemetryExporter));
+      Telemetry.setProvider(
+        Telemetry
+          .withExporter(win._dlo_telemetryExporter)
+          .withDefaultAttributes(defaultDloAttributes),
+      );
     }
+
+    const initializationSpan = Telemetry.startSpan(telemetryType.initializationSpan);
 
     if (win._dlo_observer) {
       Logger.getInstance().warn(LogMessageType.ObserverMultipleLoad);
@@ -131,7 +136,7 @@ export default function _dlo_initializeFromWindow() {
       urlValidator: win._dlo_urlValidator || undefined,
       rules,
     });
-    Logger.getInstance().record('DLO initialization time', { numericValue: startTime - Date.now() });
+    initializationSpan.end();
   } catch (err) {
     Logger.getInstance().error(LogMessageType.ObserverInitializationError, { reason: `Error: ${err}` });
   }
--- a/src/embed/init.impl.ts
+++ b/src/embed/init.impl.ts
@@ -110,6 +110,7 @@ export default function _dlo_initializeFromWindow() {
     } else {
       Telemetry.setProvider(
         Telemetry
+          // An undefined or null telemetry exporter does not send telemetry to any destination
           .withExporter(win._dlo_telemetryExporter)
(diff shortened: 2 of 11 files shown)


rollup.config.js
```diff
@@ -1,30 +1,33 @@
...
       declaration: false,
       declarationMap: false,
       sourceMap: false,
-      resolveJsonModule: false
-    })
-  ]
-};
\ No newline at end of file
+      resolveJsonModule: true,
+      moduleResolution: 'node',
```
ndonze-fs, line 30: Changing the module resolution strategy was necessary to import `package.json` via the `@rollup/plugin-json` plugin. Versioning DLO is already a little burdensome:
- `npm version`
- Add `CHANGELOG` entry
- Update version in `README`
- After PR merges, tag merged version and push new version tag

I wanted to lean on the version from `package.json` to avoid adding another step (e.g. update version in `telemetry.ts`).

src/embed/init.impl.ts
```diff
@@ -83,17 +85,17 @@ function _dlo_collectRules(): any[] {
         results.push(rule);
       });
     });
-    Logger.getInstance().record('DLO rule processing time', { numericValue: startTime - Date.now() });
+    ruleCollectionSpan.end();
     return results;
   } catch (err) {
     Logger.getInstance().error(LogMessageType.RuleRegistrationError, { reason: `Error: ${err}` });
+    Telemetry.count(telemetryType.clientError, 1, errorAttributes(errorType.ruleRegistrationError));
```
ndonze-fs, line 92: I'm interested in feedback on the overall pattern here. We could take at least a couple alternative directions:
- Add a `Telemetry.error` or similar method which handles adding the error type attribute. Would allow calling e.g. `Telemetry.error(errorType.ruleRegistrationError)`
- Report each error as a separate count and avoid the attribute altogether. We'll probably end up aggregating these, but that can be done in our custom exporter. Example if we reported each error as a separate count: `Telemetry.count(telemetryType.ruleRegistrationError, 1)`

src/handler.ts
```diff
@@ -56,6 +59,15 @@ export default class DataHandler {
    * @param event a browser Event or CustomEvent emitted
    */
   handleEvent(event: CustomEvent<DataLayerDetail>): void {
+    const handleEventSpan = Telemetry.startSpan(telemetryType.handleEventSpan, {
+      operatorCount: this.operators.length,
+      operatorNames: this.operators
```
ndonze-fs, line 64: Is this worth collecting? Figured knowing which operators are used and to what degree would be valuable. Is comma separated the best way to go? Should these at least be alphabetically sorted for consistency?

src/observer.ts
```diff
@@ -131,9 +134,18 @@ export class DataLayerObserver {
     }
 
     if (rules) {
+      const ruleRegistrationSpan = Telemetry.startSpan(telemetryType.ruleRegistrationSpan);
       rules.forEach((rule: DataLayerRule) => this.registerRule(rule));
+      // TODO(nate): Remove this call when the record log level is deprecated. Removing breaks tests
+      // so there may be some test state management issues to address
       Logger.getInstance().record('DLO rule count', { numericValue: rules.length });
```
ndonze-fs, line 141: Per the comment, removing this logging statement causes test failures. I suspect there's some global test state that isn't being handled properly. Rather than resolving in this PR which already has a decent sized scope, I plan to address this when we deprecate the `record` log level.

src/utils/telemetry.ts
```diff
@@ -1,5 +1,49 @@
...
+  ruleCount: 'dlo_rule_count',
+  handleEventSpan: 'dlo_handle_event_span',
+  clientError: 'dlo_client_error',
+};
+
+/**
+ * Constants used for error types in client error telemetry. These string values should
+ * never change
+ */
+export const errorType = {
```
ndonze-fs, line 21: I elected to have separate error constants for telemetry rather than overloading e.g. `LogMessageType.OperatorError` so these could evolve independently. Logging message types could be changed with less friction; telemetry error types should really be pinned down as exporters may implement logic based on those values.

In the current implementation, I'm also explicitly calling `Telemetry.count` after we call `Logger.getInstance().error`. We could consider always calling `Telemetry.count` within `Logger.getInstance().error`, at which point I'd probably lean toward overloading `LogMessageType` or maybe even extracting that to a dedicated `ErrorType` concept which is used by both logging and telemetry. Interested in feedback on this.

src/embed/init.impl.ts
patrick-fs: I like the `Telemetry.error` interface - usage is more explicit.

I have some questions about using the error type as the metric name: any concerns about Cardinality issues? Would using the error type this way help when reporting things in grafana (e.g. if we wanted to have a graph of all errors, visualized by type over time)?

src/embed/init.impl.ts
```diff
@@ -104,11 +106,17 @@ export default function _dlo_initializeFromWindow() {
 
     // Logging must be initialized before telemetry since telemetry errors may be logged
     if (win._dlo_telemetryProvider) {
-      Telemetry.setInstance(win._dlo_telemetryProvider);
+      Telemetry.setProvider(win._dlo_telemetryProvider);
     } else {
-      Telemetry.setInstance(Telemetry.withExporter(win._dlo_telemetryExporter));
+      Telemetry.setProvider(
+        Telemetry
+          .withExporter(win._dlo_telemetryExporter)
```
patrick-fs, line 113: Would it be a good idea to add a comment here about what the expected `Telemetry` metric collection behavior is if no `win._dlo_telemetryExporter` has been defined?

src/embed/init.impl.ts
```diff
@@ -67,7 +69,7 @@ window['_dlo_rulesFromOpsTeam'] = [
 
 function _dlo_collectRules(): any[] {
   try {
-    const startTime = Date.now();
+    const ruleCollectionSpan = Telemetry.startSpan(telemetryType.ruleCollectionSpan);
```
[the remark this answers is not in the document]
patrick-fs: I think it would be interesting to keep it in and look for any unexpected outliers in production initializations.

src/handler.ts
patrick-fs: I think insights about operators is interesting and including the operator chain in the event-handling performance metric should give us insights into whether there are any unexpected latencies in our rule sets - could be interesting for tuning default rules or understanding performance implications of custom rules.

src/handler.ts
```diff
@@ -56,6 +59,15 @@ export default class DataHandler {
...
+        .map((operator) => operator.options.name)
+        // Unique operator names
+        .filter((name, index, self) => self.indexOf(name) === index)
```
patrick-fs, line 67: Per my comment above, maintaining the order in which operators are executed (and all operators in the execution chain) could be insightful. Does this report all operators across all rules, rather than just the ones that apply to this particular event?

src/utils/telemetry.ts
patrick-fs: I think we'll need to revisit how our current logger works at some point - currently we're consuming customer FS.event quota. I can see a future where both logging and metrics are collapsed into a common framework that uses the same telemetry provider - maybe one that implements both the current otel-based interfaces as well as a "logger" interface. Idunno, just riffing; I could be way off-base about implementation.

I think the implementation as-is is fine for this PR.

src/utils/telemetry.ts
```diff
@@ -277,10 +351,31 @@ export class Telemetry {
    * Gets the configured {@link TelemetryProvider} instance. Initializes a default
    * telemetry provider if a telemetry provider hasn't been set
    */
-  static getInstance(): TelemetryProvider {
+  private static getInstance(): TelemetryProvider {
```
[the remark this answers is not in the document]
patrick-fs: I suggest sticking with SemVar - if it's a breaking change (even if no one is likely using it) then do a Major version bump.

src/utils/telemetry.ts
```diff
@@ -193,13 +239,41 @@ export class DefaultTelemetryProvider implements TelemetryProvider {
...
+   * collected telemetry events
+   *
+   * @param attributes The default attributes to include with all telemetry events
+   */
+  withDefaultAttributes(attributes: Attributes): DefaultTelemetryProvider {
+    this.defaultAttributes = attributes;
+    return this;
+  }
+
+  private insertDefaultAttributes(attributes?: Attributes): Attributes {
```
patrick-fs, line 261: What do you think about naming this `mergeDefaultAttributes`?

src/utils/telemetry.ts
```diff
@@ -193,13 +239,41 @@ export class DefaultTelemetryProvider implements TelemetryProvider {
...
+    // We're avoiding Object.assign and object literal spreading since they're
+    // not supported by IE11
+    const mergedAttributes: Attributes = {};
+    [attributes, this.defaultAttributes].forEach((attributeSet) => {
```
patrick-fs, line 265: I wonder if you could do something like:

```
const mergedAttributes = this.defaultAttributes
for (const [key, value] of Object.entries(attributes)) {
  mergedAttributes[key] = value;
}
return mergedAttribute;
```

src/embed/init.impl.ts
ndonze-fs: This is a good question. We have some precedent for tracking error type as a label with errors in Grafana, and doing so would allow us to pivot by error type in Grafana. Even if we capture error type as metric metadata here, we can also make an independent decision about whether to include error type as a label in Grafana. For example, we could drop/ignore error type or choose to log to another sink in our exporter and downstream processing.

patrick-fs: Yep - good point.

src/handler.ts
ndonze-fs: Only rules applying to the event being processed (from the defined rule which triggered the event) would be included. I could see the argument for not de-duping like you mentioned and preserving original processing order.

src/utils/telemetry.ts
ndonze-fs: This or object literal spreading would be a lot simpler. I've been trying to avoid APIs not supported by IE11 since we still have some open work to remove a few lingering IE11 errors in DLO. If we decide to no longer support IE11 in DLO or add something like babel to build process we could use `Object.assign` or object literal spreading.

src/handler.ts
patrick-fs: Ah, got it - I forgot that there's one `DataHandler` instance per rule. Yeah I think not de-duping would be better. The data is interesting and I don't think de-duping would save much in terms of payload size.

src/utils/telemetry.ts
patrick-fs: I see, I just straight missed this comment:

```
// We're avoiding Object.assign and object literal spreading since they're
// not supported by IE11
```

ndonze-fs: I'm glad you asked about this. I learned we can use object literal spreading and TypeScript (based on our compiler options) will use and polyfill `Object.assign`. We can simplify this function to:

```
  private insertDefaultAttributes(attributes?: Attributes): Attributes {
    return { ...this.defaultAttributes, ...attributes };
  }
```

When built, this results in:

```
    DefaultTelemetryProvider.prototype.insertDefaultAttributes = function (attributes) {
        return __assign(__assign({}, this.defaultAttributes), attributes);
    };

    ...

    var __assign = function() {
        __assign = Object.assign || function __assign(t) {
            for (var s, i = 1, n = arguments.length; i < n; i++) {
                s = arguments[i];
                for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p)) t[p] = s[p];
            }
            return t;
        };
        return __assign.apply(this, arguments);
    };
```

Minified build size is roughly the same, though full build size is ~1KB smaller.

ndonze-fs:
@patrick-fs This should be ready for another review. I bumped the major version and added a corresponding entry to the `CHANGELOG`. I haven't updated the `README` to point to the latest version yet. I'd prefer to wait until we release because we'll probably need to shift both the version specific and latest links to `/v2`, and these links won't be valid until we release.

src/embed/init.impl.ts
patrick-fs: Just pinging here in case you intended to convert this to a `Telemetry.error` call. Up to you.

patrick-fs (approved):
LGTM! I left one remark about `Telemetry.error` in an existing comment thread. Feel free to merge as-is, if you'd like.

PR #173, opened 2022-04-01
Exclude debounce time from measured event handling duration

ndonze-fs:
Before this change, we're including debounce time in the measured event handling duration. By default, this debounce time is 250ms, and that debounce duration will be renewed if there are additional changes to a given rule source and path within the debounce duration. While cumulative debounce time could be interesting to measure explicitly in the future, event handling duration is intended to measure the time it takes to execute an operator pipeline -- including the final function operator call to send the transformed data to the configured rule destination. This PR narrows the scope of the event handling span to operator pipeline execution (excluding debounce time).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 2.0.1
+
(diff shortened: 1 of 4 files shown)


sabrina-li (approved):
LGTM!

PR #174, opened 2022-04-11
Update mocha package version to satisfy snyk

djwhatley-fs:
We've been getting PRs from Snyk in the main FS repo to upgrade the mocha dependency. But of course we have to do that in this repo, then sync to the other one. In addition to the version change to the mocha package, I updated `@types/mocha` to the latest version too, presuming it is better to keep those in line. 

I ran the tests after installing the new version and everything seemed fine.


ndonze-fs (approved):
LGTM

PR #175, opened 2022-04-18
Add rule count as attribute to rule registration span telemetry; deprecate rule count telemetry type

ndonze-fs:
This PR adds rule count as an attribute to the rule registration span telemetry type and deprecates the dedicated rule count telemetry type for removal in a future major version release. The structure and naming of `ruleCount` matches the `operatorCount` attribute on the handle event duration telemetry type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
-### 2.1.1
+### 2.1.0
 
 - Adds rule count attribute to rule registration span telemetry. Deprecates dedicated rule count telemetry type for removal in a future release.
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 2.1.1
```
patrick-fs, line 13: Nit: I think the version will be `2.1.0`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fullstory/data-layer-observer",
-  "version": "2.0.1",
+  "version": "2.1.1",
```
patrick-fs, line 3: hmm... now that I'm seeing this here, did an `npm version` command bump from `2.0.1` to `2.1.1`?

patrick-fs (approved):
Code changes LGTM - I have a question about versioning

package.json
ndonze-fs: I did this. 🤦‍♂️ I'll get this updated to 2.1.0 before merging.

PR #177, opened 2022-04-27
Update rollup to 2.70.1 per Snyk recommendation

djwhatley-fs:
https://github.com/cowpaths/mn/pull/44773


ndonze-fs (approved):
LGTM. To test this change:
- I built DLO locally with and without this change
- The SHA256 hashes of `dlo.min.js` matched
- The SHA256 hashes of `dlo.js` (non-minified) didn't match
- As a final verification step, I ran the experimental browser tests against the new `dlo.js` version, and they all passed

PR #179, opened 2022-05-11
Fully reference latest DLO version and suggested beforeDestination operators

ndonze-fs:
Updates README to fully reference the latest DLO version (v2). Also updates `beforeDestination` operator references to match those served by FullStory data layer integrations.


sabrina-li (approved):
LGTM

PR #180, opened 2022-05-12
Add documentation for configuring telemetry measurement and exporting

ndonze-fs:


docs/telemetry.md
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+```
+
+## Customizing Telemetry Measurement
+
+If `DefaultTelemetryProvider` measurement behavior is insufficient, telemetry measurement and exporting can be entirely overridden by configuring a custom `TelemetryProvider`. By configuring a custom `TelemetryProvider` other telemetry libraries like Open Telemetry may be used for telemetry measurement and exporting.
+
+```
+class OpenTelemetryProvider {
```
ndonze-fs, line 38: Interested in thoughts about including a legit Open Telemetry example here. I think it helps illustrate how to implement a custom telemetry provider with a real world target, but also has the potential to rot as the Open Telemetry API surface evolves.

docs/telemetry.md
```diff
@@ -0,0 +1,72 @@
+# Data Layer Observer Telemetry
+
+DLO's telemetry implementation is inspired by Open Telemetry. While DLO doesn't depend on Open Telemetry, DLO telemetry can be configured to be measured by another provider like Open Telemetry or exported to a custom destination for visibility.
```
patrick-fs, line 3: Suggest linking to the Open Telemetry site

```suggestion
DLO's telemetry implementation is inspired by [Open Telemetry](https://opentelemetry.io/). While DLO doesn't depend on Open Telemetry, DLO telemetry can be configured to be measured by another provider like Open Telemetry or exported to a custom destination for visibility.
```

docs/telemetry.md
patrick-fs: I like that you're including an OTel example.  A couple of suggestions:

1. Change the title to `## Customizing Telemetry Measurement - Open Telemetry Example`
2. Include the `import` statement that defines the types and classes in your code sample:

```
import {Tracer, Meter, SimpleSpanProcessor, ...etc} from 'open-telemetry' // I don't know what the actual package name is
```

patrick-fs: Maybe this? https://www.npmjs.com/package/@opentelemetry/sdk-trace-web

ndonze-fs: Great suggestion. This forced me to reconcile changes in their metrics API surface.

ndonze-fs:
@patrick-fs This should be ready for another review.

docs/telemetry.md
patrick-fs: Nice! Thank you.

docs/telemetry.md
```diff
@@ -0,0 +1,82 @@
+# Data Layer Observer Telemetry
+
+DLO's telemetry implementation is inspired by [Open Telemetry](https://opentelemetry.io/). While DLO doesn't depend on Open Telemetry, DLO telemetry can be configured to be measured by another provider like Open Telemetry or exported to a custom destination for visibility.
+
```
patrick-fs, line 4: Might be a good idea to quickly mention that DLO can easily be configured to log to the console be default, with a link back to the `## Telemetry` header in the Readme. Good to cross-reference.

patrick-fs (approved):
LGTM! I added one little suggestion about cross-referencing between the Readme and the Telemetry docs. Please feel free to merge at will.

PR #182, opened 2022-05-13
ignore time spans that are negative or not a number

GGonryun:
This PR prevents the DefaultTelemetryProvider from exporting or sending negative or non number values.

### DIFF
--- a/test/telemetry.spec.ts
+++ b/test/telemetry.spec.ts
@@ -30,6 +30,7 @@ class MockTelemetryProvider extends MockClass {
 
 describe('DefaultTelemetryProvider', () => {
   const originalConsole = globalThis.console;
+  const originalWindow = globalThis.window;
   let mockConsole: Console;
 
   beforeEach(() => {
@@ -41,6 +42,7 @@ describe('DefaultTelemetryProvider', () => {
   });
 
   afterEach(() => {
+    (globalThis as any).window = originalWindow;
     (globalThis as any).console = originalConsole;
     Logger.getInstance().level = 1;
   });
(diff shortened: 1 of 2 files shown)


test/telemetry.spec.ts
```diff
@@ -239,6 +241,58 @@ describe('DefaultTelemetryProvider', () => {
     const countAttributes = expectParams(exporter, 'sendCount').pop().attributes;
     expect(Object.getOwnPropertyNames(countAttributes).length).to.equal(expectedAttributeCount);
   });
+
+  it('does not send span if duration is a negative number', () => {
+    const exporter = new MockTelemetryExporter();
+    const provider = new DefaultTelemetryProvider(exporter);
+
+    const span = provider.startSpan('test span');
+    provider.count('test count', 1);
```
ndonze-fs, line 250: Nit: We should be able to remove the `provider.count` call here.

test/telemetry.spec.ts
```diff
@@ -239,6 +241,58 @@ describe('DefaultTelemetryProvider', () => {
...
+
+    expectNoCalls(exporter, 'sendSpan');
+  });
+
+  it('does not send span if duration is not a number', () => {
+    const exporter = new MockTelemetryExporter();
+    const provider = new DefaultTelemetryProvider(exporter);
+
+    const span = provider.startSpan('test span');
+    provider.count('test count', 1);
```
ndonze-fs, line 267: Nit: We should be able to remove the `provider.count` call here.

test/telemetry.spec.ts
```diff
@@ -239,6 +241,58 @@ describe('DefaultTelemetryProvider', () => {
...
+
+    expectNoCalls(exporter, 'sendSpan');
+  });
+
+  it('does not send span if an error is encountered', () => {
+    const exporter = new MockTelemetryExporter();
+    const provider = new DefaultTelemetryProvider(exporter);
+
+    const span = provider.startSpan('test span');
+    provider.count('test count', 1);
```
ndonze-fs, line 284: Nit: We should be able to remove the `provider.count` call here.

ndonze-fs (approved):
Left a few suggestions, but overall LGTM.

PR #183, opened 2022-05-16
do not count rule length

GGonryun:
This PR prevents the telemetry provider from sending a `ruleRegistration` metric if there are no rules registered.

### DIFF
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -131,7 +131,7 @@ export class DataLayerObserver {
       Logger.getInstance().level = logLevel;
     }
 
-    if (rules) {
+    if (rules && rules.length > 0) {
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 2.1.1
+
+- Prevent the telemetry provider from sending a Rule Registration metric if there are no rules registered.
```
GGonryun, line 15: @ndonze-fs, i added both PR's to the changelog. lmk if the other PR should increment the version number to 2.1.2 and these should be broken up.

PR #188, opened 2022-06-16
Update CEDDL ruleset tests to be runnable via node or browsers

ndonze-fs:
This PR converts CEDDL ruleset tests which were previously only runnable via node to variants that can be run via node or Playwright-managed headless browsers (Chrome, Firefox, and Safari). The intent with this PR is to establish a pattern which can be used to update other ruleset tests to a form that can similarly be run against arbitrary DLO library artifacts in a browser.

Motivations:
- Allow running ruleset tests against arbitrary DLO transpiled artifacts (local dev, staging, prod) to build confidence in release candidates as part of the DLO dev lifecycle.
- Reuse ruleset testing logic to allow fast feedback in local dev by running ruleset tests via node while supporting greater confidence in release candidates with browser tests -- all while minimizing the potential for drift between node-based ruleset tests and browser ruleset tests.

How to run (assumes a functioning DLO local dev env):
- `npm i` to install the newly added Playwright and node types packages
- `npm run test:browser:bootstrap` to install Playwright browsers locally in `node_modules`
- `PLAYWRIGHT_DLO_SCRIPT_SRC=https://edge.fullstory.com/datalayer/v2/latest.js npm run test:browser` to run browser ruleset tests against the latest production DLO release

Overall design approach:
- Introduces a concept of test environments which may be node, Chrome, Firefox, or Safari
- Test environments know how to create test harnesses which provide implementations of functions needed by most ruleset tests (test setup, ability to inspect `FS.event` calls, ability to manipulate a data layer to trigger changes)
- Test harnesses are created and torn down once per test (for browser tests, each test harness creates its own Playwright page)
- Test environments are torn down after all tests are run (for browser tests, each test environment creates its own Playwright browser instance which is reused across all tests for that browser)


package.json
```diff
@@ -11,6 +11,8 @@
     "examples": "npm run build && http-server",
     "test": "nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts test/**/*.spec.ts",
     "test:fullstory": "nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/fullstory.spec.ts",
+    "test:browser": "PLAYWRIGHT_BROWSERS_PATH=0 nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts -- --grep 'CEDDL to FullStory'",
```
ndonze-fs, line 14: Couple notes here:
- `PLAYWRIGHT_BROWSERS_PATH=0` installs browser binaries in `node_modules/playwright-core/.local-browsers`. I took this route to isolate the impact of DLO Playwright browser tests on the broader dev environment. Interested in feedback on this approach.
- I'm expecting `--grep 'CEDDL to FullStory'` to evolve in subsequent PRs to hybridize other ruleset tests for execution in node or browser environments.

package.json
```diff
@@ -46,6 +48,7 @@
     "@types/chai": "^4.2.11",
     "@types/jsdom": "^16.2.3",
     "@types/mocha": "^9.1.0",
+    "@types/node": "^17.0.44",
```
ndonze-fs, line 51: TODO: I think we can back this change out as this was from earlier iteration which attempted to use `fs` and `path`.

test/rules-ceddl-fullstory.spec.ts
```diff
@@ -1,149 +1,159 @@
...
+          ...firstProduct,
+          productInfo: {
+            ...firstProduct.productInfo,
+            sku: 'test',
+          },
+        };
+
+        await testHarness.execute(([localFirstProduct]) => {
+          (globalThis as any).digitalData.cart.item.push(localFirstProduct);
+        }, [firstProduct]);
```
ndonze-fs, line 66: These invocations may look a little odd, but they match Playwright's `page.evaluate` calls. Whereas `node` is able to capture variables defined in the outer scope, Playwright callbacks for JavaScript evaluation in the browser require you to explicitly pass in args from the outer scope as an array. Here, we're explicitly passing in `firstProduct` from the outer scope so it can be referenced as `localFirstProduct` in the Playwright-managed browser environment.

package.json
```diff
@@ -11,6 +11,8 @@
     "examples": "npm run build && http-server",
     "test": "nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts test/**/*.spec.ts",
     "test:fullstory": "nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/fullstory.spec.ts",
+    "test:browser": "PLAYWRIGHT_BROWSERS_PATH=0 DLO_RUN_BROWSER_TESTS=1 nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts -- --grep 'CEDDL to FullStory'",
```
ndonze-fs, line 14: Couple notes here:
- `PLAYWRIGHT_BROWSERS_PATH=0` installs browser binaries in `node_modules/playwright-core/.local-browsers`. I took this route to isolate the impact of DLO Playwright browser tests on the broader dev environment. Interested in feedback on this approach.
- I'm expecting `--grep 'CEDDL to FullStory'` to evolve in subsequent PRs to hybridize other ruleset tests for execution in node or browser environments.

test/rules-ceddl-fullstory.spec.ts
```diff
@@ -1,149 +1,159 @@
...
+      });
+
+      after(async () => {
+        await testEnv.tearDown();
+      });
+
+      it('sends the first CEDDL product to FS.event', async () => {
+        await testHarness.execute(() => {
+          (globalThis as any).digitalData.product[0].attributes = { customProp: 'foo' };
+        });
```
ndonze-fs, line 32: Previous CEDDL ruleset tests specified `readOnLoad: true` which was testing a different code path (direct operator handler invocation) instead of the code path used by the CEDDL data layer integration (handler triggering via event listeners). These tests now all test dynamic data layer changes via event listeners which exercises the code paths used by the CEDDL data layer integration.

package.json
jake-b-story: On Bullet Point 1, 
I don't see any issues here, and makes sense to me. You are essentially changing the default cache folder that Playwright would normally download to (ie: `~/Library/Caches/ms-playwright` for Mac Machines).

 This might make sense to do for the `eco-test-party` as well as I imagine having multiple repositories writing to the default cache location could cause some version issues or something like that.

test/rules-ceddl-fullstory.spec.ts
```diff
@@ -1,149 +1,159 @@
...
+
+        const [eventName, payload] = await testHarness.popEvent();
```
GGonryun, line 34: is there any way for an event to timeout? can we configure that timeout limit?

test/utils/ruleset-test-harness.ts
```diff
@@ -0,0 +1,151 @@
...
+      globalThis.FS = {
+        event: (...args: any) => {
+          globalThis.events.push(args);
+        },
+      };
+
+      const dloScriptTag = document.createElement('script');
+      dloScriptTag.src = localDloScriptSrc;
+      document.body.appendChild(dloScriptTag);
+    }, [rules, dataLayer, dloScriptSrc]);
```
GGonryun, line 103: why dont we want the browser scoped rules/dataLayer/dloScriptSrc? I might be understanding this incorrectly but don't the node tests use the browser scoped rules/dataLayer/dloScriptSrc?

test/utils/ruleset-test-harness.ts
```diff
@@ -0,0 +1,151 @@
...
+    await this.page.evaluate(action, args || []);
+  }
+
+  async popEvent(timeoutMs: number = 1000) {
+    await this.page.waitForFunction(() => (window as any).events.length, undefined, { timeout: timeoutMs });
+    return this.page.evaluate(() => (window as any).events.pop());
+  }
+}
+
+export const getRulesetTestEnvironments = (): RulesetTestEnvironment[] => {
```
GGonryun, line 123: TLDR: this looks like a factory method that could use some parameters.

if my understanding is correct this is the initializer for the tests. if so we could give this method a parameter of options that will help configure these settings. 

use case: some tests for us on eco-playwright don't work on certain browsers, we may need to disable a browser/environment from running tests. i can't _currently_ imagine that would happen for DLO.
use case: as a developer, i don't want to iterate on all tests, when developing tests it might help if i can quickly specify what browsers to run, global timeout logic or settings.

not sure where this falls on overengineering vs futureproofing, wdyt?

test/utils/ruleset-test-harness.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  async setUp(rules: any, dataLayer: any) {
+    this.browser = await this.browserPromise;
+    this.page = await this.browser.newPage();
+
+    await this.page.evaluate(([localRules, localDataLayer, localDloScriptSrc]) => {
+      // This allows node and playwright tests to reference the same global/window name.
+      const globalThis: any = window;
```
GGonryun, line 84: document looks like another global variable, could we define window this way somehow, that way we don't have to stick the `:any` definition on it.

GGonryun (approved):
got a few questions but lgtm 🤔

test/utils/ruleset-test-harness.ts
```diff
@@ -0,0 +1,151 @@
...
+import {
+  Browser, Page, chromium, firefox, webkit,
+} from 'playwright';
+
+import {
+  expectFS, ExpectObserver, setupGlobals, expectGlobal,
+} from './mocha';
+
+export interface RulesetTestHarness {
+  setUp: (rules: any, dataLayer: any) => Promise<void>;
```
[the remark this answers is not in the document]
ndonze-fs: Rules should really be of type `DataLayerRule[]`. The initialization is a little weird since we're importing a JS file with the ruleset being assigned to the global `window` object -- then pulling the rules out of the `window` object typed as `any`. To increase clarity, though, I've changed `rules` within the ruleset test utils to be typed as `DataLayerRule[]`.

test/utils/ruleset-test-harness.ts
ndonze-fs: We send the browser our own local rules; however, the DLO library is loaded from the given JS file (`dloScriptSrc`) for browser tests. Here's how I think about the testing breakdown relative to the DLO dev lifecycle:

- When developing DLO locally, you'll typically run `npm test`. This executes test logic without running it through a browser, using local rulesets and local library logic (i.e. the working directory). This is a faster feedback loop and is also the test command run via CI.
- When testing DLO release candidates (e.g. a candidate built locally or deployed to a staging environment), you'll typically run `PLAYWRIGHT_DLO_SCRIPT_SRC=http... npm run test:browser`. This executes the same test logic through target headless browsers, using the same local rulesets but pulling in the DLO library logic from the given JS file (aka release candidate). The goal here is to build confidence in a transpiled JS release candidate rather than the local working directory/TS files.

Let me know what you think.

test/utils/ruleset-test-harness.ts
ndonze-fs: I think you've identified a good entry point for future config options -- whether defined here explicitly as args or via other env vars passed in on `process.env` (`DLO_RUN_BROWSER_TESTS` influences the return value here already). I'm not sure what configuration values would be useful today given we'd expect to primarily run these tests for release candidates in staging and that frequency is low. How would you feel about waiting to see what config options make sense in the future?

test/utils/ruleset-test-harness.ts
ndonze-fs: I added some global/window typings to broadly reduce `any` usage in tests and within the test env/harness logic. LMKWYT.

jake-b-story:
LGTM (I am still waiting for official approver status on FullstoryDev), so I'm just leaving the comment : )

ndonze-fs:
@GGonryun This should be ready for another look.

test/utils/ruleset-test-harness.ts
GGonryun: That makes sense, thank you for clarifying.

test/utils/ruleset-test-harness.ts
GGonryun: i'm okay with waiting 👍

GGonryun (approved):

PR #189, opened 2022-06-17
Fix node process not exiting when browser tests finish

ndonze-fs:
Following #188 merging into `main` I'm noticing the node process doesn't exit after running browser tests. I was able to narrow the cause down to the strategy implemented to reuse browser instances across tests. This PR tweaks that approach and allows the node process to exit after running browser tests.

### DIFF
--- a/test/utils/ruleset-test-harness.ts
+++ b/test/utils/ruleset-test-harness.ts
@@ -77,11 +77,9 @@ declare global {
 }
 
 class BrowserTestHarness implements RulesetTestHarness {
-  private browser: Browser = null!;
-
   private page: Page = null!;
 
-  constructor(private readonly browserPromise: Promise<Browser>) {
+  constructor(private readonly browser: Browser) {
     if (!dloScriptSrc) {
       throw new Error(
         'PLAYWRIGHT_DLO_SCRIPT_SRC must be configured with a DLO library script location for browser tests.',
@@ -90,7 +88,6 @@ class BrowserTestHarness implements RulesetTestHarness {
   }
 
   async setUp(rules: DataLayerRule[], dataLayer: any) {
-    this.browser = await this.browserPromise;
     this.page = await this.browser.newPage();
 
     await this.page.evaluate(([localRules, localDataLayer, localDloScriptSrc]) => {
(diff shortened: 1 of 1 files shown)


test/utils/ruleset-test-harness.ts
```diff
@@ -139,16 +136,17 @@ export const getRulesetTestEnvironments = (): RulesetTestEnvironment[] => {
     return [chromium, firefox, webkit].map((browserType) => {
       // Reuse a single browser instance for each browser type across all tests.
       // This reduces browser test run time by about half.
-      const browserPromise = browserType.launch();
+      let browser: Browser | undefined;
       return {
         name: browserType.name(),
         createTestHarness: async (rules: DataLayerRule[], dataLayer: any) => {
-          const testHarness = new BrowserTestHarness(browserPromise);
+          browser = browser || await browserType.launch();
```
GGonryun, line 143: If the condition were flipped we would we default to the non-nullable 'browser' object. Why does it need to be first?

ndonze-fs: I'm not sure I understand the question. The intent here is to only launch a given browser once for a test environment instance, and this leans into the fact that we're not currently parallelizing tests so the first test harness created should pay the cost. I'm okay with the compromise on parallelization since, in my testing, running these tests through Mocha with the `--parallel` switch significantly increases test run time.

GGonryun: I was wondering what the difference of "browser || browser launch" vs "browser launch || browser" in the code statement

GGonryun: This cleared it up thank you!

GGonryun (approved):

test/utils/ruleset-test-harness.ts
ndonze-fs: Ah, great, thanks! Yeah, once `browser` is defined as it'll be after the first `await` we'll no longer try to launch another browser instance for the same browser type.

GGonryun: Yeah, the scope in JS threw me off 😵

jake-b-story:
Looks good. I pulled the branch down and tested locally on my end. 

LGTM

PR #190, opened 2022-06-17
Support optionally running Google ruleset tests in browsers

ndonze-fs:
Updates Google ruleset tests (Event Measurement, Enhanced Ecommerce) to support running those tests directly in node against the working directly or via browser against a given DLO build artifact.

As part of this PR, I updated the Event Measurement ruleset to match the rules served with FullStory data layer capture integrations. I left the Enhanced Ecommerce ruleset as is. While this ruleset has also diverged, reconciling it will be more work and may be better for a dedicated scope of work.


package.json
```diff
@@ -11,7 +11,7 @@
     "examples": "npm run build && http-server",
     "test": "nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts test/**/*.spec.ts",
     "test:fullstory": "nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/fullstory.spec.ts",
-    "test:browser": "PLAYWRIGHT_BROWSERS_PATH=0 DLO_RUN_BROWSER_TESTS=1 nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts -- --grep 'CEDDL to FullStory'",
+    "test:browser": "PLAYWRIGHT_BROWSERS_PATH=0 DLO_RUN_BROWSER_TESTS=1 nyc mocha -r ts-node/register -r source-map-support/register -r jsdom-global/register test/*.spec.ts -- --grep Ruleset --timeout 3000",
```
ndonze-fs, line 14: I change the grep pattern to capture all converted ruleset tests and increased the Mocha test timeout from 2000ms to 3000ms. I haven't commonly exceeded this test timeout, but I've occasionally run into this locally when browser instances haven't launched expediently.

rulesets/google-event-measurement.js
```diff
@@ -1,65 +1,12 @@
-window['_dlo_rules_google_measurement'] = [
+window["_dlo_rules_google_em"] = [
```
ndonze-fs, line 1: I updated this ruleset to match the ruleset being served with FullStory data layer capture integrations.

test/rules-ceddl-fullstory.spec.ts
```diff
@@ -14,7 +14,7 @@ declare global {
   var digitalData: CEDDL;
 }
 
-describe('CEDDL to FullStory rules', () => {
+describe('Ruleset: CEDDL to FullStory', () => {
```
ndonze-fs, line 17: Name change to enable a test grep pattern to include all converted ruleset tests (at this point, CEDDL and Google).

test/rules-google-enhanced-ecommerce-fullstory.spec.ts
```diff
@@ -0,0 +1,343 @@
+import 'mocha';
+
+import { expectEqual } from './utils/mocha';
+import { RulesetTestHarness, getRulesetTestEnvironments } from './utils/ruleset-test-harness';
+
+// TODO(nate): Update this ruleset to match the rules served by FullStory data layer capture integrations.
+// This will require supplying a "convert" beforeDestination operator to convert string currency values
+// to numbers.
+import '../rulesets/google-ua-enhanced-ecommerce.js';
```
ndonze-fs, line 9: I opted not to update the Enhanced Ecommerce ruleset to match the ruleset served with FullStory data layer capture integrations in this PR. This ruleset is more complicated, and the changes cut deeper and would've resulted in a larger PR surface. I'm hoping to true up this ruleset in a future dedicated PR.

test/rules-google-fullstory.spec.ts
```diff
@@ -1,448 +0,0 @@
-import { expect } from 'chai';
```
ndonze-fs, line 1: I split tests for the 2 Google rulesets (Enhanced Ecommerce and Event Measurement) into separate files. These are separate integrations/config options in FullStory and are captured as distinct rulesets in the DLO repo, so it felt natural to break them apart into separate files.

test/rules-google-enhanced-ecommerce-fullstory.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+                    category: 'product gallery',
+                    variant: '',
+                  },
+                ],
+              },
+            },
+          });
+        });
+
+        let [eventName, payload] = await testHarness.popEvent();
```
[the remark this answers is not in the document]
jake-b-story: I might be missing the point. But if FS event calls like a stack, then popEvent() seems like an accurate name to me since a stack has the option to 'pop' once populated. 

Dequeue would imply FS event calls are treated more like a queue, where we could remove an event from either end of the list.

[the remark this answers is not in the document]
ndonze-fs: > I might be a little confused, but if FS event calls like a stack, then popEvent() seems like an accurate name to me since a stack has the option to 'pop' once populated.

This is the current behavior, and I agree with the current behavior `popEvent()` is an appropriate name. I'm proposing that it might be clearer if we treat FS event calls like a queue instead of a stack and instead pluck events FIFO -- in which case `dequeueFSEventCall()` would seemingly be a more appropriate name.

Since we're in control of how the `FS.event` calls are treated in the test harness, we have the option of treating the calls as either a stack or a queue. I guess it's worth considering that both test harnesses (node and browser) would need updating, and the node test harness is currently leaning on a lot of prior `expect*` logic which may be more burdensome to unwind than I'd anticipated. Maybe it's worth deferring this change?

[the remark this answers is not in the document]
jake-b-story: Thanks for clearing this up, I realize now I didn't read your first comment carefully enough. 

I'd say it's your choice if you want to defer the change or not. It makes sense to me either way, so I have no strong opinion if the change is in this PR or some other one.

PR #193, opened 2022-06-22
Support optionally running custom ruleset tests in browsers

ndonze-fs:
This file tests edge cases that aren't covered by rulesets served with data layer capture integrations.

### DIFF
--- a/test/rules-custom.spec.ts
+++ b/test/rules-custom.spec.ts
@@ -1,72 +1,87 @@
 import 'mocha';
 
 import { DataLayerRule } from '../src/observer';
-import {
-  expectEqual, waitForFS, setupGlobals, ExpectObserver, expectGlobal, expectNoCalls,
-} from './utils/mocha';
+import { expectEqual } from './utils/mocha';
+import { RulesetTestHarness, getRulesetTestEnvironments } from './utils/ruleset-test-harness';
+import { AppMeasurement } from './mocks/adobe';
 
-describe('Custom rules', () => {
-  afterEach(() => {
-    ExpectObserver.getInstance().cleanup();
-  });
+declare global {
+  // eslint-disable-next-line no-var, vars-on-top
+  var s: AppMeasurement;
+}
 
-  it('should only trigger rules for property changes that match rule selector', async function test() {
-    // Extend test timeout to accommodate waits for event dispatching
-    (this as any).timeout(3000);
+describe('Ruleset: Custom rules', () => {
+  getRulesetTestEnvironments().forEach((testEnv) => {
+    describe(`test environment: ${testEnv.name}`, () => {
+      let testHarness: RulesetTestHarness;
 
-    const s = {
-      eVar11: 'test',
-      eVar22: 'test',
-    };
+      afterEach(async () => {
+        await testHarness.tearDown();
+      });
 
-    setupGlobals(([
-      ['s', s],
-    ]));
+      after(async () => {
+        await testEnv.tearDown();
+      });
 
-    const rules: DataLayerRule[] = [
-      {
-        source: 's[^(eVar1)]',
-        operators: [
+      it('only triggers rules for property changes that match rule selector', async () => {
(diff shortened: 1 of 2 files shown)


test/utils/mocha.ts
```diff
@@ -298,31 +298,3 @@ export function expectFS(methodName: FSMethodName, namespace = 'FS'): any[] {
...
-
-/**
- * Similar to `expectFS`, `waitForFS` is a convenicne method for `expectParams` that
- * checks the global FullStory object. Unlike `expectFS`, `waitForFS` supports waiting
- * for a global FullStory object method call within a timeout period.
- * @param methodName FullStory API function (`methodName` arg for `expectParams`)
- * @param namespace Global object if FullStory is not `FS`
- * @param timeout Time in milliseconds after which the Promise is rejected
- */
-export function waitForFS(methodName: FSMethodName, namespace = 'FS', timeout: number = 500): Promise<any[]> {
```
ndonze-fs, line 310: Converting these tests to be runnable via browsers removed the only consumer of this function.

test/rules-custom.spec.ts
```diff
@@ -1,72 +1,87 @@
...
-    ExpectObserver.getInstance().create({ rules });
+        testHarness = await testEnv.createTestHarness(rules, {
+          s: {
+            eVar11: 'test',
+            eVar22: 'test',
+          },
+        });
 
-    expectGlobal('s').eVar11 = 11;
+        await testHarness.execute(() => {
```
GGonryun, line 57: this test technically doesn't follow AAA pattern, i do think that's okay here but the way these tests are organized makes it look like two separate tests in the AAA pattern. no action to be taken, just a brain dump.

ndonze-fs: True; we have more of an Arrange -> Act -> Assert -> Act -> Assert flow here. I find myself leaning into flows like this more in integration test scenarios where multiple actions build upon each other to demonstrate the target behavior. In this case, we previously would've triggered both rules when editing a property matching either rule source.

We could update this test to perform both actions at once (setting `eVar11` and `eVar12`) -- then reversing our assertions since the `FS.event` call queue is currently treated like a stack. LMK if you think that would be clearer or if you have other ideas for how to make this test less surprising while also clearly verifying the intended behavior -- setting a property value matching one rule source doesn't trigger another rule on the same base object (`s` in this case) whose source doesn't match the changed property.

test/rules-custom.spec.ts
```diff
@@ -1,72 +1,87 @@
...
-    await new Promise<void>((resolve) => {
-      setTimeout(() => {
-        expectNoCalls(expectGlobal('FS'), 'event');
-        resolve();
-      }, 1000);
+        // Verify no other calls were made to rule destinations
+        await new Promise<void>((resolve, reject) => {
+          testHarness.popEvent(500)
+            .then(() => {
+              reject(new Error('Expected rejected promise due to no FS.event calls being present.'));
```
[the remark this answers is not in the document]
ndonze-fs: Agreed; I sort of regret not refactoring this earlier. Fortunately, this should be the last PR before getting out a general browser test cleanup PR.

I was originally thinking of introducing something like `await testHarness.ensureNoFSEventCalls()` that would reject a promise like we have here, but your comment makes me wonder if it would be better to update `testHarness.popEvent()` to return `undefined` if the timeout is reached instead of rejecting the promise. This would let us:

```
expect(await testHarness.popEvent(500)).to.be.undefined;
```

WDYT?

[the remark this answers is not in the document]
GGonryun: I like the idea of returning undefined here instead, it makes the expect look really neat 👍 


this doesn't require action and is more for my understanding: does a rejected promise throw an exception and if so could we catch it? sample code:

```
try {
  await testHarness.popEvent(500)
  expect.fail()
} catch (exception) {
   expect(exception).is.notNull
}
```

PR #194, opened 2022-06-23
Cleans up recent browser test efforts

ndonze-fs:
- Extracts common logic waiting to verify no `FS.event` calls are queued
- Fixes a race condition that was resolving differently in CI than locally

### DIFF
--- a/test/utils/ruleset-test-harness.ts
+++ b/test/utils/ruleset-test-harness.ts
@@ -14,7 +14,7 @@ export interface RulesetTestHarness {
   setUp: (rules: DataLayerRule[], dataLayer: any) => Promise<void>;
   tearDown: () => Promise<void>;
   execute: (action: (args: any[]) => void, args?: any[]) => Promise<void>;
-  popEvent: (timeoutMs?: number) => Promise<any[]>;
+  popEvent: (timeoutMs?: number) => Promise<any>;
 }
 
 interface RulesetTestEnvironment {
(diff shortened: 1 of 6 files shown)


test/utils/ruleset-test-harness.ts
```diff
@@ -45,14 +45,14 @@ const nodeTestHarness: RulesetTestHarness = {
...
-  popEvent: async (timeoutMs: number = 1000) => new Promise<any[]>((resolve, reject) => {
+  popEvent: async (timeoutMs: number = 1000) => new Promise<any>((resolve) => {
     const startTime = new Date().getTime();
 
     const checkForEvents = setInterval(() => {
       const currentTime = new Date().getTime();
       if (currentTime - startTime >= timeoutMs) {
         clearInterval(checkForEvents);
-        reject(new Error(`Timeout ${timeoutMs} exceeded waiting for FS.event call.`));
+        resolve(undefined);
```
ndonze-fs, line 55: Instead of rejecting the promise which throws when awaited, it seemed nicer to return `undefined` which could be asserted against for tests verifying no `FS.event` calls were queued.

GGonryun (approved):
LGTM!

GGonryun:
>I spiked reusing browser instances across all ruleset tests. This reduced local test run time from 51s down to 42s

I was expecting a bigger difference, but i think the +9 seconds is alright here too.

jake-b-story (approved):
LGTM

PR #196, opened 2022-06-24
Match test rulesets to those served by data layer integrations

ndonze-fs:
We did some of this work along the way to hybridizing ruleset tests to be runnable in node or browsers. This PR finishes work to true up rulesets by:
- Matching the Google Enhanced Ecommerce ruleset to the rules served with the related data layer integration
- Adding Adobe App Measurement and Tealium Retail to documented rulesets


test/rules-google-enhanced-ecommerce-fullstory.spec.ts
```diff
@@ -68,7 +68,7 @@ describe('Ruleset: Google Analytics Enhanced Ecommerce to FullStory', () => {
         let [eventName, payload] = await testHarness.popEvent();
         expectEqual(eventName, 'detail_product');
         expectEqual(payload.id, 'P000525722');
-        expectEqual(payload.price, 2.99);
+        expectEqual(payload.price, '2.99');
```
ndonze-fs, line 71: In the previous ruleset version in this repository, a convert operator converted these numeric strings to numbers. For data layer integrations, that conversion is handled by a `beforeDestination` operator. Since we're testing the ruleset in isolation, these values aren't converted. In the future, it may be worth testing these rulesets + expected `beforeDestination` operators to more completely test data layer integration behavior.

rulesets/google-ua-enhanced-ecommerce.js
```diff
@@ -1,325 +1,156 @@
...
-        "value": "pageview"
-      }
+      { "name": "query", "select": "$[?(pageType, pageName)]" },
+      { "name": "insert", "value": "pageview" }
     ],
     "destination": "FS.event"
   },
   {
-    "id": "fs-ua-e-commerce-detail-action",
+    "id": "fs-ga-e-commerce-detail-action",
```
GGonryun, line 11: does the name of the file still relevant if we're removing all the ua references?

GGonryun (approved):
LGTM, just one question about the file name for my understanding 🤔

PR #197, opened 2022-06-28
Null guard for selectors

van-fs:
It appears that Google Analytics events can contain `null` objects.  This is supposed to "clear" Google from applying previously seen data to subsequent events.  We've seen this occur in two forms:
- `dataLayer.push(null);`
- `dataLayer.push({"ecommerce":null{);`

The problem this presents is that our rulesets for GA are looking into these values trying to use the query operator to tell what type of event it is.  Since the value is null, reading a property results in an exception.  For example, a GA EC rule is doing, `{ "name": "query", "select": "$.ecommerce.detail.actionField" }` which causes an error because there is no `detail` key.  DLO silently fails because it catches the error; however, DLO will emit log events that a failure occurred.  This means that N FS.events are created for each of the N rules.  This is noisy and likely to trigger rate limiting.

I've added guards to the `selector.ts` as well as unit tests for `selector` and higher level ones for the `QueryOperator` and Google rulesets.


test/rules-google-enhanced-ecommerce-fullstory.spec.ts
```diff
@@ -335,6 +335,17 @@ describe('Ruleset: Google Analytics Enhanced Ecommerce to FullStory', () => {
...
+
+      it('gracefully handles a "clear" event', async () => {
+        await testHarness.execute(() => {
+          globalThis.dataLayer.push({
+            ecommerce: null,
+          });
+        });
+
+        const event = await testHarness.popEvent();
+        expectEqual(event, undefined);
```
ndonze-fs, line 347: Nit: Elsewhere when looking for no events to be triggered, we're:
- Setting a timeout of 500ms via `await testHarness.popEvent(500);`. In local testing usually 250ms is enough time to trigger any events so we've been doubling that to verify that no events were triggered.
- Using `expect(event).to.be.undefined;`; do you have a preference for which form to standardize on?

test/rules-google-event-measurement-fullstory.spec.ts
```diff
@@ -61,6 +61,17 @@ describe('Ruleset: Google Analytics Event Measurement to FullStory', () => {
...
       });
+
+      it('gracefully handles a "clear" event', async () => {
+        await testHarness.execute(() => {
+          globalThis.dataLayer.push(
+            null,
+          );
+        });
+
+        const event = await testHarness.popEvent();
```
ndonze-fs, line 72: Nit: Similar suggestion of potentially constraining to a 500ms timeout via `testHarness.popEvent(500);`. This keeps our test feedback loop a little shorter.

As I type this, I'm wondering about reducing our default timeout to 500ms (currently 1000ms). I'm happy to take that on as a TODO in a separate scope of work.

test/rules-google-enhanced-ecommerce-fullstory.spec.ts
```diff
@@ -335,6 +335,17 @@ describe('Ruleset: Google Analytics Enhanced Ecommerce to FullStory', () => {
         expectEqual(eventName, 'refund');
         expectEqual(payload.id, 'T12345');
       });
+
+      it('gracefully handles a "clear" event', async () => {
```
ndonze-fs, line 339: I would've expected this test to fail against the latest DLO production version, but it doesn't:

```
$ npm i && npm run test:browser:bootstrap
$ PLAYWRIGHT_DLO_SCRIPT_SRC=https://edge.fullstory.com/datalayer/v2/latest.js npm run test:browser -- --grep Google
```

Is there something else we should be asserting against? What's the current behavior in production without these changes? I'm happy to help with any test changes here since I recently introduced test environments and test harnesses to ruleset tests.

test/selector.spec.ts
```diff
@@ -259,4 +260,26 @@ describe('test selection paths', () => {
   it('filter notation should return a reference to the object', () => {
     expect(select('favorites.films[?(action=Rogue One)]', testData)).to.eq(testData.favorites.films);
   });
+
+  it('filter notation should return a reference to the object', () => {
+    expect(select('favorites.films[?(action=Rogue One)]', testData)).to.eq(testData.favorites.films);
+  });
```
ndonze-fs, line 266: Nit: Looks like this test case was duplicated somehow (exists immediately above).

ndonze-fs (commented):
Left some small questions and comments related to tests. Let me know what you think.

test/selector.spec.ts
van-fs: Yep, ty; appears that I copied/pasted intending to use and then didn't :(

test/rules-google-enhanced-ecommerce-fullstory.spec.ts
van-fs: - The `500` ms timeout makes sense.  The `DataHandler.DefaultDebounceTime` is `250` ms so I'd expect that to be the floor of time needed to wait.  The need to wait applies only to shimmed property observation, which is why I also added the unit tests to `selector` and`QueryOperator` since those are synchronous.
- We've used `.to.be.undefined` in the past.  I went with `expectEqual` because that helper function was added to remove ambiguity from doing shallow versus deep comparison.  There are 4 equality tests in mocha, and in most comparisons of DLO, we want deep comparisons since we should be making copies of objects from the data layer.  So I'd probably say `expectEqual` since we'd want to use it as a best practice.

test/rules-google-event-measurement-fullstory.spec.ts
van-fs: Sure, or something you might use is `1.5 * DataHandler.DefaultDebounceTime` (`1.5` is illustrative, use `2` for comfort).  Not sure if we can access the class member `debounce` of `DataHandler` from tests, but that would be the more accurate debounce timeout.

test/rules-google-enhanced-ecommerce-fullstory.spec.ts
van-fs: Ah, yes.  The undesirable side effect is that the `FullStoryAppender` creates FS.events when the `ERROR` level is applied.  This is noisy and triggers rate limiting.  (Updated the original PR description with these details.)

So I suppose that a better test is that the data layer event is undefined AND no log event was created as a result.  In other parts of the code, we do `expectNoCalls(globalMock.console, 'log');`.  Is that still valid for the test harness?  Or does the test harness allow for a log sink that is scoped to each test case?  (The latter seems needed.)

van-fs: @ndonze-fs Sure, push it and I'll review.

ndonze-fs: @van-fs As I'm working through this change locally, I'm running into an issue with the GA Event Measurement tests logging an operator error against the latest revision in this PR. You should be able to see this error logged in the test output locally by either running all Google Analytics tests:

```
$ npm test -- --grep 'Google Analytics'
```

...or running just the Event Measurement tests:

```
$ npm test -- --grep 'Google Analytics Event Measurement'
```

Output:

> Operator Error {"operator":"query","path":"dataLayer.push","reason":"Cannot convert undefined or null to object"}

Any ideas? This error is causing test failures when I start checking for errors and assert there are none.

van-fs: I see it now.  With the original problem is addressed, the `isEmptyData` function will throw because it's trying to look at keys for a null value.  Investigating the behavior.  This code was desired to prevent "empty" data from being sent to FullStory IIRC.

test/rules-google-event-measurement-fullstory.spec.ts
```diff
@@ -61,6 +61,20 @@ describe('Ruleset: Google Analytics Event Measurement to FullStory', () => {
...
+        await testHarness.execute(() => {
+          globalThis.dataLayer.push(
+            null,
+          );
+        });
+
+        const event = await testHarness.popEvent(500);
+        expectEqual(event, undefined);
+
+        const error = await testHarness.popError(500);
```
ndonze-fs, line 75: Current status: This fails if running only Google Analytics ruleset tests and passes when running the entire test suite. Investigating if we have some global state carryover that's causing the difference.

ndonze-fs: When running the full test suite the log appender had been changed away from `ConsoleAppender`. This could arguably be missed cleanup in other tests, but it also seemed reasonable that since that Node test harness is specifically inspecting `console` calls for errors that it ensures the `ConsoleAppender` is active -- so I've taken that approach.

ndonze-fs:
@irv-fs Since Van and I both contributed changes to this PR, we could use another perspective.

irv-fs (approved):
This looks good to me.  If you want me to give more attention to anything specifically that you are having a discussion on, let me know.

ndonze-fs (approved):
@van-fs This LGTM if our combined changes look good to you.

van-fs:
@ndonze-fs LGTM.  I updated the branch.  Ready for your approval.

ndonze-fs (approved):
@van-fs LGTM. Feel free to merge away.

PR #198, opened 2022-07-13
Adding in ignore attribute for convert operator to be paired with enumerate

irv-fs:


van-fs:
cc @ndonze-fs for awareness and co-approval

src/operators/convert.ts
```diff
@@ -104,7 +108,14 @@ export class ConvertOperator implements Operator {
 
     // if enumerate is set, try to coerce all strings into an equivalent numeric value
     if (enumerate) {
-      const enumerableProps = ConvertOperator.enumerableProperties(data[index]);
+      let enumerableProps = ConvertOperator.enumerableProperties(data[index]);
+      // if ignore properties are set, make sure those are filtered out
+      if (ignore) {
+        /* no matter what I did, breaking it up into multiple lines, or using curly braces, I couldn't get the line
```
van-fs, line 114: Ideally I'd prefer if we could remove the comments related to linting.  Perhaps we could just do `const filterPredicate = (key) => ignore.find((element:string) => element === key) === undefined);` and then `enumerableProps.filter(filterPredicate)`.  (Note code not testing, just typing it here for example.)

irv-fs: I can remove the comments about the linting.  I even broke up the lines into separate inner functions, and it still complained, and would reformat the code on me.  I spent 30 minutes on it and gave up.  I'll remove the comment and see if your suggestion will fix it.

irv-fs: That worked and I updated the pull request to remove the comment and the ignore line.

src/operators/convert.ts
```diff
@@ -104,7 +108,12 @@ export class ConvertOperator implements Operator {
 
     // if enumerate is set, try to coerce all strings into an equivalent numeric value
     if (enumerate) {
-      const enumerableProps = ConvertOperator.enumerableProperties(data[index]);
+      let enumerableProps = ConvertOperator.enumerableProperties(data[index]);
+      // if ignore properties are set, make sure those are filtered out
+      if (ignore) {
```
ndonze-fs, line 113: I'm noticing `ignore` is typed as `any` here. Would it make sense to add this property to `ConvertOperatorOptions` e.g.:

```
export interface ConvertOperatorOptions extends OperatorOptions {
  // ...
  ignore?: string | string[];
  // ...
}
```

src/operators/convert.ts
```diff
@@ -104,7 +108,12 @@ export class ConvertOperator implements Operator {
...
+        const filterPredicate = (key:string) => ignore.find((element:string) => element === key) === undefined;
```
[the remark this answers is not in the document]
irv-fs: That is way easier to read and much better.  I put that in!

src/operators/convert.ts
irv-fs: I totally missed this.  Good catch!

ndonze-fs (approved):
LGTM!

PR #199, opened 2022-07-18
Remove record log level

ndonze-fs:
The `record` log level was originally positioned to collect stats but was never implemented. Instead, in DLO v2.0.0 [telemetry was enhanced to collect stats](https://github.com/fullstorydev/fullstory-data-layer-observer/blob/main/CHANGELOG.md#200). Since we're approaching a DLO major version bump, this PR removes the `record` log level which has always been a noop but which could be referenced by library consumers. Accordingly, it seems to make sense to treat this as a breaking change.

### DIFF
--- a/test/observer.spec.ts
+++ b/test/observer.spec.ts
@@ -691,10 +691,6 @@ describe('DataLayerObserver unit tests', () => {
     expect(observer.handlers.length).to.eq(1);
     expect(globalMock.dataLayer.length).to.eq(0);
 
-    // Ignore the first two as they are observation messages
-    expectParams(appender, 'log');
-    expectParams(appender, 'log');
-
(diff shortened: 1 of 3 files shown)


test/observer.spec.ts
```diff
@@ -691,10 +691,6 @@ describe('DataLayerObserver unit tests', () => {
     expect(observer.handlers.length).to.eq(1);
     expect(globalMock.dataLayer.length).to.eq(0);
 
-    // Ignore the first two as they are observation messages
-    expectParams(appender, 'log');
-    expectParams(appender, 'log');
```
ndonze-fs, line 696: These 2 ignored calls were for `record` level log entries which have been removed.

PR #204, opened 2022-08-26
CEDDL digitalData.cart.item rule is no longer readOnLoad: false

ndonze-fs:
Reflects changes to default, out-of-the-box rules.

### DIFF
--- a/test/rules-ceddl-fullstory.spec.ts
+++ b/test/rules-ceddl-fullstory.spec.ts
@@ -1,5 +1,4 @@
 import 'mocha';
-import { expect } from 'chai';
 
 import { expectEqual, expectMatch, expectUndefined } from './utils/mocha';
 import { RulesetTestHarness, getRulesetTestEnvironments } from './utils/ruleset-test-harness';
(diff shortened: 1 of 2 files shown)


test/rules-ceddl-fullstory.spec.ts
```diff
@@ -86,12 +85,6 @@ describe('Ruleset: CEDDL to FullStory', () => {
         expectUndefined(payload, 'linkedProduct');
       });
 
-      it('does not send CEDDL cart item products to FS.event on load', async () => {
```
ndonze-fs, line 89: This test didn't fail after removing `readOnLoad: false` from the CEDDL ruleset. It turns out this test stopped verifying the target behavior when I hybridized ruleset tests to run in a Node env or via Playwright browser automation. For this test to have been verifying the target behavior, we would've needed to specify a global `readOnLoad: true` configuration to be overridden by the rule's `readOnLoad: false`.

Either way, I think it makes sense to remove this test now since there's no longer an exceptional behavior to verify.

sabrina-li (approved):

PR #206, opened 2022-11-29
README Update for readOnLoad Default

liz-fs:
The default for readOnLoad was changed as a result of this ticket: https://app.shortcut.com/fullstory/story/203569/update-dlc-ootb-integrations-to-have-dlo-readonload-set-to-true-by-default


van-fs:
@liz-fs I think the DLO library itself still defaults to false.  The FullStory integration will configure the `_dlo_readOnLoad` window variable to always be `true`, but absent that, DLO will default to `false` according to this [line](https://github.com/fullstorydev/fullstory-data-layer-observer/blob/da2e4e4eaa0b24413904bc243bcd9dd2a702e2e9/src/embed/init.impl.ts#L138).  LMK if I'm missing something.

PR #208, opened 2022-12-19
Ignore properties that are already suffixed

irv-fs:


src/operators/suffix.ts
```diff
@@ -139,6 +139,23 @@ export class SuffixOperator implements Operator {
     }
   }
 
+  /**
+   * Check to see if a prop already contains a suffix
+   * @param prop The prop to check
+   * @return If the prop has a suffix on it already
+   */
+  static isAlreadySuffixed(prop: string) : boolean {
+    if (prop === undefined || prop === null) {
```
van-fs, line 148: What do you think about a quick test if the char at index N has an `_`?  Since we know the suffixes, we could perform a direct access test to see if a suffix looks to be present.  If so, then we do the subsequent test (the loop) to verify if a valid suffix is present.

van-fs:
For posterity, @irv-fs can you add a comment on why this change is being made?  Is it to prevent double-suffixing (i.e. `someProperty_str_str`)?

src/operators/suffix.ts
irv-fs: We could do that.  We'd have to check for length - 3 and length - 4 for the various sizes.  If someone were to add a longer suffix, we'd have to update the code.  I tried to make it so it was more generic, just working off the enum.

irv-fs:
> 

I wasn't sure where you were talking about.   If you are talking about where i passed the suffixed values into the unit test, it was just a convenient way to test all of the suffixes, as the test data has all the suffixes.  So it was a quick way to test that they all work and get ignored.

van-fs:
@irv-fs I meant the PR itself.  There was chatter about the `ConvertOperator` and how it leads to `_str` versus `_strs`.  But that's an unintended side effect of default configuration, not suffix operation.  I just want to ensure we capture the problem and solution statement if we have to look back on the commit history.

src/operators/suffix.ts
```diff
@@ -139,6 +139,23 @@ export class SuffixOperator implements Operator {
...
+      return false;
+    }
+    let matches = false;
+    const suffixes:Array<Suffixes> = Object.values(Suffixes);
```
ndonze-fs, line 152: It seems like we could move this outside the function scope and call `Object.values` once instead of once for every property we evaluate. I'm not sure if we'll run into friction with the current `static` modifier on this function, though.

If we're only calling `Object.values` once, we could also consider reducing these values to an object with keys matching known type suffixes. That'd give us way to check a substring against object keys. I'm not sure if performance is a concern here or which route would be more performant, but capturing the last underscore index and checking any property suffix against an object with keys containing known suffixes is how the FullStory Segment cloud mode integration checks for known type suffixes. This also allows you to short-circuit the O(n) `endsWith` check against all known type suffixes if there isn't an underscore as @van-fs was suggesting. Examples from our open source Segment cloud mode integration:

- [Type suffix validators containing underscore-less suffixes as keys](https://github.com/segmentio/action-destinations/blob/3356398b24c12eda10495b75e73afa72d6c1968f/packages/destination-actions/src/destinations/fullstory/vars.ts#L37-L52)
- [Check for known type suffixes](https://github.com/segmentio/action-destinations/blob/3356398b24c12eda10495b75e73afa72d6c1968f/packages/destination-actions/src/destinations/fullstory/vars.ts#L98-L100)

Feel free to take or leave any of these ideas, but extracting the `Object.values` call may be worth considering if that change doesn't introduce too much friction.

src/operators/suffix.ts
van-fs: Agree that new suffixes would require a change.  I'm thinking that a property with an existing suffix is low probability since this data is coming from the data layer and would likely never have a FullStory-specific suffix.  So the quick test would efficiently check before doing a more exhaustive check of all possible suffixes.

ndonze-fs (approved):
LGTM. Left a suggestion to consider, but I don't see that as a blocker to merging the existing solution.

src/operators/suffix.ts
irv-fs: I can put that in if folks want.  With library code that can live a long time, I usually err on the side of not putting in potential stumbling blocks, and hard coding a position looking for suffixes that could change if someone adds a suffix is against my nature.  Not a hill for me to die on, so if you folks want it, I can put that in.  endsWith is not a super expensive operator

src/operators/suffix.ts
irv-fs: I'll move the Object.keys into a variable at the same level as Suffixes so we don't have to do it over and over.  I don't think it is a huge performance gain, but it is an easy fix.

irv-fs:
This PR was put in as the FullStory usage of DLO has a _dlo_beforeDestination that includes a convert operator and then a suffix operator.  This combination means that some values, such as IDs, that contain all numeric characters were being converted to a number.  There is a way to override _dlo_beforeDestination, but is requires a full override.  This mechanism would allow suffixing the ID property inside the individual rule and it would then be ignored by the built in suffix operator.

src/operators/suffix.ts
irv-fs: Also in looking at our suffixes, we have some that are 3, 4 and 5 in length.  To do a quick test to see if they have the _ in the right position, I'd have to look for it at all three of those positions.  I could look for the presence of a _ character, but not sure that is any faster than endsWith.  In the examples posted, it looks like the split the value by the _ character, but I'm not sure that is any faster.  If something isn't much faster, I'd vote for readability.  In this case, it is a suffix operator, which means a string endsWith.

ndonze-fs: New suffixes shouldn't cause an issue, and suffix length shouldn't be a problem either if you snag a single substring based on the last index of `_` and check that against an object with keys containing known suffix. This is what our Segment cloud mode destination and our FS client APIs do (e.g. `FS.setUserVars`). https://github.com/fullstorydev/fullstory-data-layer-observer/pull/208#discussion_r1054530767

irv-fs:
Yes I saw that.  I can change the code to work that way if it is
preferred.  Not a huge deal to me either way.

irv-fs:
I updated the code to use a lookup, which also has the added benefit of a
quick switch out if there is no _ character, hopefully satisfying both
concerns.

ndonze-fs:
> Yes I saw that. I can change the code to work that way if it is preferred. Not a huge deal to me either way.

I'm good with this solution as is or with these proposed changes. However, I'd propose mirroring the approach from our Segment cloud mode integration and the FS client api if we're interested in short circuiting the suffix check in a way that would gracefully handle new suffixes and suffixes of different lengths.

I'd speculate the approach used in the Segment cloud mode integration and the FS client api would be more performant, but we'd want to benchmark that if we want to justify the change for performance reasons. I don't think the approach used in the Segment cloud mode integration or FS client api is less readable, maintainable, or heavier in terms of LOC, though.

src/operators/suffix.ts
```diff
@@ -139,6 +146,23 @@ export class SuffixOperator implements Operator {
...
+   */
+  static isAlreadySuffixed(prop: string) : boolean {
+    if (prop === undefined || prop === null) {
+      return false;
+    }
+    const lastIndex = prop.lastIndexOf('_');
+    if (lastIndex < 0) {
+      return false;
+    }
+    const trialSuffix = prop.substring(lastIndex, prop.length);
```
ndonze-fs, line 162: Nit: You shouldn't need a second arg to `substring in this case.

src/operators/suffix.ts
```diff
@@ -30,6 +30,13 @@ export enum Suffixes {
   Reals = '_reals',
 }
 
+// squirrel the suffix values away so we can use it as a lookup later
+const SuffixValueLookup = new Map<string, string>();
```
ndonze-fs, line 34: We could also use an `Object` as a `Record<string, string>` (or `Record<string, boolean>` since the value isn't significant). I don't have a strong argument for changing this away from a `Map`, though. If we used an object, a common pattern would be to reduce the `Object.values` e.g.:

```
const suffixValueLookup = Object.values(Suffixes).reduce<Record<string, boolean>>(
    (acc, val) => ({
      ...acc,
      [val]: true
    }),
    {}
  );
```

No change required here unless you prefer this approach.

ndonze-fs (approved):
LGTM. Left a couple small things to consider, but nothing required to merge this change.

src/operators/suffix.ts
irv-fs: I went to change it to this, but I also lost the has method, which seems like it became less readable.  I did get rid of a temporary variable to look more like this but left the Map.  Researched online and it didn't seem to be a major difference in performance

irv-fs:
@van-fs Pointed out to me that if you pass in a property and put a suffix on it that doesn't match the actual type, we threw an  error.  That was unexpected (I was thinking it would just honor the suffix). Therefore, my overall fix won't work unless we also fix the `convert` operator.  I added a new flag `ignoreSuffixed` that will ignore suffixed properties.  The combination of that plus this suffix operator fix will mean most `_dlo_beforeDestination` fixes for improper conversion will not be needed, saving a lot of effort and hassle for our customers.  This does mean we will need to default ignoreSuffixed in `_dlo_beforeDestination` to be true so that this takes effect.

van-fs:
@irv-fs Do we need the `ignoreSuffixed` flag?  It seems like this change is a more desirable behavior: honor the suffix, assume the data is typed correctly, and do not convert.  If `convert` doesn't ignore a suffixed property, we have the possibility that someone applies conversion to an already suffixed property leading to a mismatch (unless they follow up with a `rename` to add the correct suffix).

It seems simpler to say, "If you have a suffixed property, convert ignores it and assumes that a developer knew what they were doing."  And if there's a need to override, we can use `rename` rather than adjusting flags or altering Go code to flip the flag.

irv-fs:
If we were starting from scratch I would not have a ignoreSuffixed flag, as it is the more desirable behavior.  I was worried about breaking rules that were out there, so this flag makes it an opt in situation versus retro fitting.

docs/operator_convert.md
```diff
@@ -8,15 +8,16 @@ Convert is most useful when paired with `FS.event` for cart and checkout events
...
+|------------------|------------------------|-------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `enumerate`*     | `boolean`              | `false`     | Automatically converts all string values into their numeric equivalent.                                                                                                                                             |
+| `ignore`         | `string` or `string[]` | `undefined` | Can be paired with `enumerate` option to ignore one or more properties when performing the `enumerate`.                                                                                                             |
+| `ignoreSuffixed` | `boolean` | `true`      | Can be paired with `enumerate` option to ignore any properties that are suffixed when performing the `enumerate`.  [Suffixes](https://help.fullstory.com/hc/en-us/articles/360020623234#property-name-requirements) |
```
ndonze-fs, line 15:
```suggestion
| `ignoreSuffixed` | `boolean` | `true`      | Can be paired with `enumerate` option to ignore any properties with [known type suffixes](https://help.fullstory.com/hc/en-us/articles/360020623234#property-name-requirements) when performing the `enumerate`. |
```

docs/operator_convert.md
```diff
@@ -159,6 +160,50 @@ Options with an asterisk are required. Note that `enumerate` can be used by itse
...
 ```
+## Automatically convert all "numeric" strings ignoring suffixed property
+
+### Rule
+
+```javascript
+{
+ source: 'digitalData.cart.price[(userId, available,basePrice,cartTotal,priceWithTax)]',
+ operators: [
+   { name: 'convert', enumerate: true, ignoreSuffixed: true },
```
ndonze-fs, line 171: I guess we could omit `ignoreSuffixed` now that ignoring suffixes is the default.

ndonze-fs (approved):
I left a few small suggestions, but otherwise LGTM.

docs/operator_convert.md
irv-fs: > LGTM. Left a couple small things to consider, but nothing required to merge this change.

Made those changes.

PR #212, opened 2023-04-17
Update version to 4.0.0

djwhatley-fs:


CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ Data Layer Observer follows semantic versioning when releasing updates.
 
 ## History
 
+### 4.0.0
+
+- CEDDL `digitalData.cart.item` rule no longer has `readOnLoad` set to false.
```
ndonze-fs, line 15: I'm wondering if we can omit this change since it doesn't impact library functionality (only updates an example). WDYT?

djwhatley-fs: Oh, sure. It wasn't clear to me what that was, and I assumed that was the breaking change you had mentioned. I think we might need a better way of communicating what things from which PRs should go in the changelog.

djwhatley-fs: Also, looking at this changelog, I don't understand what the breaking change is. I guess it's related to the ignore suffix option? But as it's written here that doesn't seem like a breaking change.

ndonze-fs: Yeah, the ignore suffix option on the convert operator is the breaking change -- specifically that it defaults to `true`. This should be a relatively safe change, though it's technically a non-backward-compatible behavior change.

djwhatley-fs: Cool - I've added a note that the option defaults to true.

PR #222, opened 2023-08-11
add ends with operator for properties

liz-fs:


test/selector.spec.ts
```diff
@@ -235,6 +235,10 @@ describe('test selection paths', () => {
     expect(select('favorites[?(number+=25)]', testData)).to.be.undefined;
     expect(select('gtmEvent[?(event!^gtm.)]', testData)).to.be.undefined;
     expect(select('gtmEvent[?(event=^gtm.)]', testData)).to.not.be.undefined;
+    expect(select('gtmEvent[?(event=$gtm)]', testData)).to.be.undefined;
+    expect(select('gtmEvent[?(event=$dom)]', testData)).to.not.be.undefined;
```
van-fs, line 239: L239 is failing to pass.  LMK if you're having difficulty tracking this down.

van-fs (approved):
LGTM; @liz-fs do you have a PR to bump the version and update the CHANGELOG?

src/utils/object.ts
```diff
@@ -53,3 +53,23 @@ export function startsWith(target: string, searchString: string, position?: numb
...
 }
+
+/**
+ * Returns true if the searchString sequence is the same as the corresponding target sequence
+ * starting at the given target position (defaulting to 0); otherwise returns false.
+ * @param target The string to search within.
+ * @param searchString The string to search for.
+ * @param position The position to start searching witin the target.
+ */
+export function endsWith(target: string, searchString: string, position?: number): boolean {
```
ndonze-fs, line 64: Since we're attempting to match ES2015 `String.prototype.endsWith` behavior, how would you feel about adding tests for this function to [object-utils.spec.ts](https://github.com/fullstorydev/fullstory-data-layer-observer/blob/272309ffb01fa7874d5b271bdb34eac09bc0ef99/test/object-utils.spec.ts) to verify the behavior matches? Tests for our local `startsWith` implementation should provide a good pattern.

src/utils/object.ts
```diff
@@ -53,3 +53,23 @@ export function startsWith(target: string, searchString: string, position?: numb
 
   return target.indexOf(searchString, effectivePosition) === effectivePosition;
 }
+
+/**
+ * Returns true if the searchString sequence is the same as the corresponding target sequence
+ * starting at the given target position (defaulting to 0); otherwise returns false.
+ * @param target The string to search within.
+ * @param searchString The string to search for.
+ * @param position The position to start searching witin the target.
```
ndonze-fs, line 62: I wonder if we should update the documentation for this `@param` to closer align with ES2015 `endsWith` params. The ES2015 `endsWith` function accepts an end position, whereas here we're describing this param as a start position (possibly a copy/paste leftover from the our local `startsWith` function). Could use MDN docs for inspiration: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

src/utils/object.ts
```diff
@@ -53,3 +53,23 @@ export function startsWith(target: string, searchString: string, position?: numb
 
   return target.indexOf(searchString, effectivePosition) === effectivePosition;
 }
+
+/**
+ * Returns true if the searchString sequence is the same as the corresponding target sequence
+ * starting at the given target position (defaulting to 0); otherwise returns false.
```
ndonze-fs, line 59: Similar to updating the `@param` doc for `position`, it may make sense to update the reference here from "starting" to "ending" position as well as the mention of the default value of `0` to the default of the given `target` string length.

README.md
```diff
@@ -179,7 +179,7 @@ The `source` property uses a custom selector syntax.  It’s most often seen as
 | object[^(property, ...)] | `digitalData.cart.price[^(shipping)]` | Returns an object with properties whose names begin with `property`. |
 | object[$(property, ...)] | `digitalData.page.pageInfo[$(Date)]` | Returns an object with properties whose names end with `property`. |
 | object[?(property, ...)] | `digitalData.cart[?(cartID)]` | Returns the object or null if the object does not have property. |
-| object[?(property=value, ...)] | `digitalData.cart.price[?(basePrice>=10)]` | Returns the object or null if the object's `property` does not compare to `value`. Comparison can be `=`, `<=`, `>=`, `<`, `>`, `!=`, `!^`, and `=^`. Use the value `undefined` to query if a property `=` or `!=` undefined. |
+| object[?(property=value, ...)] | `digitalData.cart.price[?(basePrice>=10)]` | Returns the object or null if the object's `property` does not compare to `value`. Comparison can be `=`, `<=`, `>=`, `<`, `>`, `!=`, `=$`, `!$`, `!^`, and `=^`. Use the value `undefined` to query if a property `=` or `!=` undefined. |
```
ndonze-fs, line 182: Nit: In the example ordering we have `=$` and then `!$` for ends with, but `!^` and then `=^` for starts with (reversing the example order). It may make sense to be consistent with ordering between ends with and starts with operators.

Not necessary under this PR scope, but since we're accumulating a long list of example operators I wonder if it makes sense to break these out into their own table in the future -- each row with their own description - for easier consumption.

src/utils/object.ts
liz-fs: updated the comment to accurately reflect the logic behind the function!

src/utils/object.ts
liz-fs: updated the params to accurely reflect the usage of the params!

PR #223, opened 2023-09-06
update changelog and version

liz-fs:
just updating changelog in conjunction with https://github.com/fullstorydev/fullstory-data-layer-observer/pull/222


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fullstory/data-layer-observer",
-  "version": "4.0.1",
+  "version": "4.0.2",
```
ndonze-fs, line 3: Was `npm version` used to bump this? I would've expected `package-lock.json` to be updated as well.

ndonze-fs:
This pattern has ongoing friction, but we also have a link to the latest version by version number in the readme. Can we get this updated as well? https://github.com/fullstorydev/fullstory-data-layer-observer/blob/main/README.md?plain=1#L58

liz-fs (commented):
@ndonze-fs All updates regarding versioning have been made!

ndonze-fs (approved):

PR #225, opened 2023-09-19
Add GA4 compatible rulesets and tests

ndonze-fs:
This PR:

- Adds GA4 compatible Event Measurement and Enhanced Ecommerce rulesets
- Adds tests for those new rulesets


test/rules-google-event-measurement-ga4-fullstory.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            }],
+          );
+        });
+
+        const event = await testHarness.popEvent(500);
+
+        expect(event).to.be.undefined;
+      });
+
+      it('gracefully handles a "clear" event', async () => {
```
ndonze-fs, line 84: Is this what we'd expect a gtag clear event to look like?

rulesets/google-enhanced-ecommerce-ga4.js
```diff
@@ -0,0 +1,167 @@
+window['_dlo_rules_google_ec_ga4'] = [
+  {
+    "id": "fs-ga4-e-commerce-select_item",
+    "source": "dataLayer",
+    "operators": [
+      { "name": "query", "select": "$[?(event=select_item)]" },
+      { "name": "query", "select": "$.ecommerce.items[0]" },
```
ndonze-fs, line 7: Could `ecommerce.items` ever be empty? Is this a test case we should capture?

rulesets/google-enhanced-ecommerce-ga4.js
```diff
@@ -0,0 +1,167 @@
...
+    ],
+    "destination": "FS.event"
+  },
+  {
+    "id": "fs-gtg4-e-commerce-select_item",
+    "source": "dataLayer",
+    "operators": [
+      { "name": "query", "select": "$[(0,1,2)]" },
+      { "name": "query", "select": "$[?(1=select_item)]" },
+      { "name": "query", "select": "$[2].items[0]" },
```
ndonze-fs, line 88: Could `items` ever be empty? Is this a test case we should capture?

test/rules-google-enhanced-ecommerce-ga4-fullstory.spec.ts
```diff
@@ -0,0 +1,616 @@
...
+            },
+          });
+        });
+
+        const [eventName, eventProps] = await testHarness.popEvent();
+        expectEqual(eventName, 'select_item');
+        expectEqual(eventProps.item_id, 'sku_123');
+        expectEqual(eventProps.item_name, 'first item');
+        expectEqual(eventProps.price, 1.23);
+        expect(eventProps.item_list_name).to.be.undefined;
```
ndonze-fs, line 72: Would we want to capture root level event properties like `item_list_name`? This can also be defined on individual `items`.

test/rules-google-enhanced-ecommerce-ga4-fullstory.spec.ts
```diff
@@ -0,0 +1,616 @@
...
+                  item_name: 'first item',
+                  price: '1.23',
+                },
+                {
+                  item_id: 'sku_456',
+                  item_name: 'second item',
+                  price: '4.56',
+                },
+              ],
+              currency: 'USD',
```
ndonze-fs, line 137: Would we want to capture root level properties like `currency`?

test/rules-google-enhanced-ecommerce-ga4-fullstory.spec.ts
```diff
@@ -0,0 +1,616 @@
...
+        expectEqual(firstProduct.item_name, 'first item');
+        // TODO(nate): We may need to convert these string values to real
+        expectEqual(firstProduct.price, '1.23');
+        expectEqual(firstProduct.quantity, 1);
+
+        const secondProduct = eventProps.products[1];
+        expectEqual(secondProduct.item_id, 'sku_456');
+        expectEqual(secondProduct.item_name, 'second item');
+        // TODO(nate): We may need to convert these string values to real
+        expectEqual(secondProduct.price, '4.56');
```
ndonze-fs, line 559: Is there a way for us to run the `convert` operator against `items` properties?

rulesets/google-enhanced-ecommerce-ga4.js
PerStirpes: I believe items can be an empty string `""`

rulesets/google-enhanced-ecommerce-ga4.js
van-fs: I think we should.  Two outcomes, either 1) the query operator causes the chain to exit because the selected object is undefined or 2) we could add an operator step before to check that the index 0's value is not undefined.

But I have a feeling the isEmptyData check will cause the chain to exit.
```
// if the data is null or empty, it is a signal to stop processing
        // this can happen if an upstream handler needed to prevent a downstream operator
        // and we don't want to submit empty events to the configured destination
        if (handledData === null || DataHandler.isEmptyData(handledData)) {
          this.runDebugger(`[${i}] ${name} halted`, handledData, '  ');
          return null;
        }
```

rulesets/google-enhanced-ecommerce-ga4.js
van-fs: @PerStirpes Can you link to any supporting documentation or evidence from a live website?

test/rules-google-event-measurement-ga4-fullstory.spec.ts
van-fs: I've not found corresponding doc similar to [this](https://developers.google.com/analytics/devguides/collection/ua/gtm/enhanced-ecommerce#clear-ecommerce).  for standard event measurement, I think testing a `null` push to the data layer makes a lot sense to verify we're not creating an event like you've done.

rulesets/google-enhanced-ecommerce-ga4.js
ndonze-fs: It looks like we gracefully handle empty `items` arrays. I added tests to verify all events except `purchase` result in:

- 0 events
- 0 errors

irv-fs (approved):
LGTM thanks!

rulesets/google-enhanced-ecommerce-ga4.js
ndonze-fs: Going to resolve this as we have a test that verifies empty array behavior. [Enhanced Ecommerce specifies](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtm#purchase) that `items` is an array and is required. It feels like we're covering the spec well and can adapt if we get feedback from real world use.

test/rules-google-enhanced-ecommerce-ga4-fullstory.spec.ts
ndonze-fs: Capturing the first `items` element and root level properties would require a new operator which is out of scope but could be considered if we get feedback that root level properties are valuable.

PR #226, opened 2023-10-12
Fix Google ruleset references / add GA4 references

ndonze-fs:
Updates the ruleset `README`:

- Fixes broken Google ruleset reference links
- Adds GA4 reference links


djwhatley-fs (approved):
LGTM

PR #227, opened 2023-11-06
Fix GA4 browser ruleset tests to pass parent scope values

ndonze-fs:
These tests were passing when running in a Node environment. However, when passing in a browser parent scope values must be explicitly passed to the test harness. See Adobe tests for an example: https://github.com/fullstorydev/fullstory-data-layer-observer/blob/main/test/rules-adobe-fullstory.spec.ts#L43-L51


djwhatley-fs (approved):
LGTM

PR #232, opened 2024-04-04
Removes some syntax incompatible with ie11

ScottLNorvell:


src/operators/suffix.ts
```diff
@@ -1,3 +1,4 @@
+import '../typeAugmentations/allow';
```
ScottLNorvell, line 1: This is a pattern we use in fs.js. Ie11 supports maps, but technically es5 does not. See workaround here.

sabrina-li: curious: is this to make typescript happy? (`Map` sin't showing any errors for me, but it seems like it should since we only have `es5` libs for ts?)

test/operator-convert.spec.ts
```diff
@@ -308,7 +308,7 @@ describe('convert operator unit tests', () => {
   });
 
   it('strings can be enumerated', () => {
-    const range = [...Array(101).keys()];
+    const range = Array.from({ length: 101 }).map((_, i) => i);
```
sabrina-li, line 311: Looks like `Array.from` is also not supported in IE11? (this is in test so probably fine but maybe it should be tested in IE1..?)

src/operators/suffix.ts
```diff
@@ -32,7 +33,8 @@ export enum Suffixes {
 
 // squirrel the suffix values away so we can use it as a lookup later
 const SuffixValueLookup = new Map<string, string>();
-Object.values(Suffixes).forEach((item) => {
+Object.keys(Suffixes).forEach((key) => {
+  const item = Suffixes[key as keyof typeof Suffixes];
```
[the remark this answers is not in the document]
ScottLNorvell: Yeah, could even just be an array lookup since it's not that many keys. Then we could skip the Type augmentation. 🤔 

I was trying to touch as little as possible here 😄

test/operator-convert.spec.ts
ScottLNorvell: I think I'll just make this test a for loop instead of building an array and `forEach`ing it

src/operators/suffix.ts
ScottLNorvell: Yeah I was getting an error in my editor for `Map`. (Didn't try using the build system 🤔 ) 

I'll see if we can remove or refactor.

src/operators/suffix.ts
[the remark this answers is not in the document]
ScottLNorvell: Using `Set` and creating a `SuffixKey` type.

test/operator-convert.spec.ts
ScottLNorvell: Just used a for loop

success

tsconfig.json
```diff
@@ -1,12 +1,11 @@
 {
   "compilerOptions": {
-    "target": "es2015",
+    "target": "ES5",
     "module": "commonjs",
     "declaration": true,
     "declarationMap": true,
     "sourceMap": true,
     "strict": true,
-    "keyofStringsOnly": true,
```
ScottLNorvell, line 9: field was deprecated, my editor didn't like it: https://www.typescriptlang.org/tsconfig#keyofStringsOnly

ScottLNorvell:
Comments addressed @sabrina-li! PTAL

tsconfig.json
```diff
@@ -1,12 +1,11 @@
 {
   "compilerOptions": {
-    "target": "es2015",
+    "target": "ES5",
     "module": "commonjs",
```
sabrina-li, line 4: hmm the rollup config has the typescript plugin.. Admitted I'm not exactly sure what it does or if they needs to be synced..

sabrina-li: actually nvm, I think this is correct, the rollup already uses `ES5` for target.

sabrina-li (approved):
LGTM!

tsconfig.json
ScottLNorvell: yeah, it was somewhat unclear how the two of those interacted 😂 . But at least the tsconfig causes `Object.values` to error out so that's a win 🙌

PR #234, opened 2024-04-08
bump build deps

ScottLNorvell:
Webkit browser test was failing and this playwright bump fixed it success

~~Also ran `npm audit fix` to clean up some bad build deps.~~

Actually that broke CI so this is only the playwright bump

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 executors:
   browser-executor:
     docker:
-      - image: circleci/node:20-browsers
+      - image: cimg/node:20g-browsers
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 executors:
   browser-executor:
     docker:
-      - image: circleci/node:12-browsers
+      - image: cimg/node:16.20-browsers
```
ScottLNorvell, line 5: necessary for new version of playwright 

(webkit tests weren't succeeding)

sabrina-li (approved):
LGTM!

PR #235, opened 2024-12-10
New fsApi / version 2 support

irv-fs:


src/observer.ts
```diff
@@ -9,6 +9,10 @@ import { FunctionOperator } from './operators';
 import DataLayerTarget from './target';
 import MonitorFactory from './monitor-factory';
 import { errorType, Telemetry, telemetryType } from './utils/telemetry';
+import TrackEventOperator from './operators/fsApi/trackEvent';
```
van-fs, line 12: nit: these can be exported from the `index.ts` in `/operators` and imported as a single statement.

src/observer.ts
```diff
@@ -60,14 +64,23 @@ export interface DataLayerRule {
...
+  destination?: string | Function;
   readOnLoad?: boolean;
   url?: string;
   id?: string;
   description?: string;
   monitor?: boolean;
   waitUntil?: number | Function;
   maxRetry?: number;
+  version?: number;
+  fsApi?: string;
```
van-fs, line 76: `fsApi` is one of `FS_API_CONSTANTS`, no?  Should that be the type and not `string`.

src/observer.ts
```diff
@@ -200,16 +217,36 @@ export class DataLayerObserver {
...
         beforeOptions.forEach((operator) => handler.push(this.getOperator(operator)));
       }
 
-      // end with destination
-      const func = previewMode ? previewDestination : destination;
-      handler.push(new FunctionOperator({ name: 'function', func }));
+      if (fsApi) {
+        switch (fsApi) {
+          case FS_API_CONSTANTS.SET_IDENTITY:
+            handler.push(new SetIdentityOperator({ name: 'setIdentity' }));
```
van-fs, line 230: We could just reuse the `FS_API_CONSTANTS` member as the value of `name`?

src/observer.ts
```diff
@@ -190,8 +203,12 @@ export class DataLayerObserver {
    * adding the operator, the DataHandler will be removed to prevent unexpected data processing.
    * @param handler to add operators to
    * @param options for operators used to configure each Operator
+   * @param destination The javascript function to execute (must be one of destination or fsApi)
+   * @param fsApi The special FullStory constant to be executed (must be one of destination or fsApi)
    */
-  private addOperators(handler: DataHandler, options: OperatorOptions[], destination: string | Function) {
+  private addOperators(handler: DataHandler, options: OperatorOptions[],
+    destination: string | Function | undefined = undefined, fsApi: string | undefined = undefined,
```
van-fs, line 210: `destination` and `fsApi` can be declared optional with `destination?` and `fsApi?`. Is there a need to add the type `undefined` and assign it `undefined`?  That would be automatic with the optional.

src/observer.ts
```diff
@@ -417,13 +460,37 @@ export class DataLayerObserver {
...
+    // sanity check destination and fsApi parameters
+    if (!destination && !fsApi) {
+      Logger.getInstance().error(LogMessageType.OperatorError, {
+        reason: LogMessage.MissingDestination,
+      });
+      Telemetry.error(errorType.operatorError);
+      return;
+    }
+
+    if (destination && fsApi) {
```
van-fs, line 479: While conceptually incorrect (competing destinations), I wonder if we should prefer fsApi if it is also used with destination. In this scenario, we still still send data versus nothing.  We could emit an error to indicate the problem, and the user would be able to see the event in the replay along with the indication there are competing destinations.

src/operators/fsApi/fsApi.ts
```diff
@@ -0,0 +1,42 @@
...
+  static specification = {};
+
+  constructor(public options:FSApiOperatorOptions) {
+    // sets this.options
+  }
+
+  handleData(data: any[]): any[] | null {
+    const thisArg: object = getGlobal();
+    // @ts-ignore
+    const fsFunction:any = thisArg[thisArg._fs_namespace]; // eslint-disable-line
```
van-fs, line 18: I wonder if there's a way to use the typings from `@fullstory/browser`. This way we get type safety from the official library.

src/observer.ts
```diff
@@ -200,16 +217,36 @@ export class DataLayerObserver {
...
-      if (beforeDestination) {
+      if (beforeDestination && (version === 1)) {
         const beforeOptions = Array.isArray(beforeDestination) ? beforeDestination : [beforeDestination];
         beforeOptions.forEach((operator) => handler.push(this.getOperator(operator)));
       }
 
-      // end with destination
-      const func = previewMode ? previewDestination : destination;
-      handler.push(new FunctionOperator({ name: 'function', func }));
+      if (fsApi) {
```
van-fs, line 227: Given these are operators, my skew is that we re-use the `OperatorFactory` and `this.getOperator` to instantiate operators.  I may have to get into the code to see if that creates unwanted complexity.

src/operators/fsApi/setIdentity.ts
```diff
@@ -0,0 +1,20 @@
+import FSApiOperator from './fsApi';
+
+export default class SetIdentityOperator extends FSApiOperator {
+  // eslint-disable-next-line class-methods-use-this
+  prepareData(inputData:any[]): any[] | null {
+    if (inputData === null || inputData.length < 1) {
+      return null;
+    }
+    const realData:any[] = [
```
van-fs, line 9: I think we can consolidate and skip checking the `inputData`. For example, we do `const [uid, properties] = inputData`. And then later use `{uid,properties}` as the second element in `realData`.  This assumes `properties` is allowed to be undefined in the FS API.

src/operators/fsApi/setPageProperties.ts
```diff
@@ -0,0 +1,17 @@
+import FSApiOperator from './fsApi';
+
+export default class SetPagePropertiesOperator extends FSApiOperator {
+  // eslint-disable-next-line class-methods-use-this
+  prepareData(inputData:any[]): any[] | null {
+    if (inputData === null || inputData.length < 1) {
```
van-fs, line 6: Should we throw an `Error`?  The operator expects inputData, right?  Is there a scenario where not receiving data but expecting it to be sent to `setProperties` is OK?  Returning `null` is typically used to halt the operator chain under expected conditions - like when a query doesn't match and the chain is expected to stop.

src/operators/fsApi/setIdentity.ts
van-fs: Or could we do:
```    return [
      'setIdentity',
      { name: inputData[0],
        properties: inputData[1],
       },
    ];
```
As is done elsewhere in the new operators.

ScottLNorvell (commented):
Hi @irv-fs ! Thanks for your patience on this. Everything _looks_ good here. I'm going to go schedule some time to go over this with @ndonze-fs to make sure I understand what's going on and then we'll stamp

src/observer.ts
irv-fs: Fixed

src/observer.ts
irv-fs: fixed

src/observer.ts
irv-fs: I started with this, but optional parameters have to go at the end, and version is optional but with a hardcoded default value.  I'd have to move version in front of them, and then anyone using the api would have to put a version in if they wanted to specify destination or fsApi.  The other option would be to give these default values of undefined as well.  Thoughts?

src/observer.ts
irv-fs: I wanted it explicit in the code, but could move those into registered operators with an OperatorFactory if desired.  I tend to favor explicit code in special situations like this, but not a hill for me to die on.

src/observer.ts
irv-fs: Fixed

src/observer.ts
irv-fs: I tend to think if someone specified both it was probably done in error, and I'd rather bomb out than have them hunt down why it isn't working.  This happens with DLO unfortunately and it can be tricky to figure out why things are broken.  But not a hill for me to die on either, so could change it.

src/operators/fsApi/setPageProperties.ts
irv-fs: Fixed

src/observer.ts
```diff
@@ -200,16 +217,36 @@ export class DataLayerObserver {
         handler.push(this.getOperator(optionSet));
       });
 
-      // optionally perform a final transformation
+      // optionally perform a final transformation if version is 1
       // useful if every rule needs the same operator run before the destination
-      if (beforeDestination) {
+      if (beforeDestination && (version === 1)) {
```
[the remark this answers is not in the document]
irv-fs: Yes, the way this is coded, if you use V2 of DLC you are in charge of putting all of your operators that are needed, which would include the convert operator.  I do automatically insert dlo as an additional parameter in the fsApi base operator, so that piece is taken care of.  You can use the new fsApi stuff with version 1 however if desired.

src/observer.ts
```diff
@@ -190,8 +203,12 @@ export class DataLayerObserver {
    * adding the operator, the DataHandler will be removed to prevent unexpected data processing.
    * @param handler to add operators to
    * @param options for operators used to configure each Operator
+   * @param destination The javascript function to execute (must be one of destination or fsApi)
+   * @param fsApi The special FullStory constant to be executed (must be one of destination or fsApi)
    */
-  private addOperators(handler: DataHandler, options: OperatorOptions[], destination: string | Function) {
+  private addOperators(handler: DataHandler, options: OperatorOptions[],
+    destination: string | Function | undefined = undefined, fsApi: string | undefined = undefined,
+    version: number = 1) {
```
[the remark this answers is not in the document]
irv-fs: Added jsDoc.  That was an oversight.

src/observer.ts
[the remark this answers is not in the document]
irv-fs: I was very nervous to just make the suffix operator a no-op.  I know we have done a lot of work to get around suffixes, allowing people to use specific names.  I instead decided to introduce a rule version, which I think gives us more flexibility.  Anybody who wants this new functionality needs to explicitly set the rule version, which they knowingly opt into.  We can also use this for future backwards incompatible changes at the rule level.  I think it gives us another lever that was missing in the past.

ScottLNorvell (approved):
This all looks good from my end. I think once @ndonze-fs signs off we can get this deployed. 

Sorry for the delay on this.

src/observer.ts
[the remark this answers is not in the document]
irv-fs: When a customer moves to V2, it won't be as easy as flipping a switch or setting a simple attribute as each rule will have to be visited as some inserts, etc will need to be removed.  The structure of V2 is just different, so it won't be a simple path regardless. Forcing them to also consider convert seems small compared to what is already needed, and gets us away from the hardcoded beforeDestination stuff while keeping backwards compatibility.

src/observer.ts
```diff
@@ -417,13 +459,37 @@ export class DataLayerObserver {
...
+    if (!source) {
       Logger.getInstance().error(LogMessageType.RuleInvalid,
-        { rule: id, source, reason: `Missing ${source ? 'destination' : 'source'}` });
+        { rule: id, source, reason: 'Missing source' });
       Telemetry.error(errorType.invalidRuleError);
       return;
     }
 
+    // sanity check destination and fsApi parameters
+    if (!destination && !fsApi) {
```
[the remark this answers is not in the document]
irv-fs: In my design doc I had originally positioned the V2 as just a special string inside a destination, but it seemed more folks wanted the separate attribute (which I also prefer as it is clearer the intention).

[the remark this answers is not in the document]
ndonze-fs: In this case, we wouldn't have a special string; rather, we would have a more complex destination object that wouldn't look terribly different than operator definitions. That said, this is not a hill I will die on if other folks think it's sensible to start merging the ideas of operators/transformations and data sinks. 😆

src/observer.ts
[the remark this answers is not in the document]
ndonze-fs: Can you provide examples of other concerns we'll have moving customers from V1 to V2? You mentioned the structure of V2 being different which suggests that -- even if we were to position the V2 API as a destination instead of an operator -- we wouldn't be able to take an existing operator chain and have it call V2 APIs without capture behavior changes.

I'm pulling on this thread because in the past we've seen capture behavior changes (e.g. when we've changed `beforeDestinationOperators`) as breaking changes which motivated us to:
- Bump the DLO major version
- Feature switch such that we could migrate customers between major versions over time

In this case, it's a little different because we're introducing a rule version within the overall DLO library version.

Between data sinks becoming operators and opting into breaking changes via rule specs, we're shifting responsibilities around vs. where they've lived in the past. This may be okay, though I think we should have a good awareness of these shifts as they may complicate what changes we can make and where over time e.g. not being able to count on `beforeDestinationOperators` to make changes with client integration scripts since we can no longer count on those to be effective for all customers.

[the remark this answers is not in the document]
ndonze-fs: Looking back on an earlier comment, is this true?

> You can use the new fsApi stuff with version 1 however if desired.

With FS API calls currently being positioned as operators instead of destinations, wouldn't the operator to call FS API V2 execute before the `beforeDestinationOperators` which are appended to the end of the operators array?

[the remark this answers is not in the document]
irv-fs: Opening up a rule version would allow those writing rules to gradually migrate as they see fit, even rule by rule, versus an all or nothing change, and doesn't require any major hassles on our end (as far as migrating folks). Given that most of the rule writers are inside Fullstory, this seems to be a benefit?

ndonze-fs (approved):
LGTM

PR #236, opened 2025-01-30
Scottnorvell/update changelog etc

ScottLNorvell:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fullstory/data-layer-observer",
-  "version": "4.0.4",
+  "version": "5.0.0",
```
ndonze-fs, line 3: 🤔 Do we need a major version bump? I think all breaking changes (no type suffixes, ignoring `beforeDestination`) are opt-in at the rule definition level.

ScottLNorvell: Yeah wasn't sure. I can make it minor release.

irv-fs (approved):
LGTM

package.json
ndonze-fs: cc @irv-fs to confirm that we should be good with a minor version bump

ScottLNorvell: Talked to @irv-fs He is good with minor.

PR #240, opened 2025-07-29
Irv fs/convert nested

irv-fs:
This is to add the ability for the `convert` operator to work on nested objects.  Right now it only works on one level, this adds a new property called `maxDepth` (following the pattern of the `flatten` operator), that triggers the new behavior.


jurassix (approved):
LGTM

PR #243, opened 2025-12-23
Adding in honoring preview mode with fsApi and a unit test for that.

irv-fs:
I realized when coding some changes to Companion to support fsApi that previewMode was not honored with fsApi set.  This probably won't happen too much in production, but if it did it would confuse folks.  So I fixed it and added a test for it so we don't miss something like that again.


RyanCommits (approved):
LGTM, TY @irv-fs !

PR #244, opened 2025-12-23
feedback

RyanCommits:
What do you think about this change? I think it's more clear and straightforward than overriding the variables.

### DIFF
--- a/src/observer.ts
+++ b/src/observer.ts
@@ -231,30 +231,34 @@ export class DataLayerObserver {
         beforeOptions.forEach((operator) => handler.push(this.getOperator(operator)));
       }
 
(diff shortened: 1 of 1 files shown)


irv-fs:
That duplicates some logic which I am not a big fan of, as it creates duplicate places for folks to have to change things for future enhancements.  (The version > 1, inserting dlo as an additional parameter). If we really want to rework this code, we can probably move all the logic for previewMode up outside the if( fsApi )....  I typically try to rework the code as little as possible as so many important customers rely on this.

irv-fs:
I made some changes to not have to reassign variables and not to duplicate any logic.

PR #245, opened 2026-01-21
Adding in cookie source

irv-fs:


src/utils/cookie.ts
```diff
@@ -0,0 +1,28 @@
+// eslint-disable-next-line import/prefer-default-export
+export function getAvailableCookies() {
+  const cookies : Map<string, string> = new Map<string, string>();
```
jurassix, line 3: wish there was a simple way for us to know from the DLO context which browser api's are compatible with fsjs. This one is so no concern but having to look it up vs having it be a type error is error prone.

jurassix (approved):
LGTM

PR #247, opened 2026-04-27
[VAL-10160] Restart functionality

ScottLNorvell:
Makes DLO aware aware of the fs.js capture lifecycle such that it stops and restarts based on the `FS('observe')` api. https://developer.fullstory.com/browser/fullcapture/callbacks-and-delegates/


ScottLNorvell:
@cursor review

src/observer.ts
```diff
@@ -345,6 +348,10 @@ export class DataLayerObserver {
     debounce = DataHandler.DefaultDebounceTime,
     version:number = 1,
   ): DataHandler {
+    if (this.destroyed) {
+      throw new Error('DataLayerObserver has been destroyed');
+    }
+
```
ScottLNorvell, line 357: Originally we were throwing here. I updated it to just log so it wouldn't break but if anyone is expecting the Handler, it might break. (Not sure if this is a thing)

Anyway, they would need to already have a reference to DLO since we are clearing it out on shutdown anyway.

src/embed/init.impl.ts
```diff
@@ -94,6 +99,80 @@ function _dlo_collectRules(): any[] {
...
+  const ns = win._fs_namespace;
+  const fs = (typeof ns === 'string' && ns) ? win[ns] : undefined;
+  if (typeof fs !== 'function') {
+    return;
+  }
+
+  let hadFsShutdown = false;
+
+  fs('observe', {
+    type: 'shutdown',
```
eugene-chang-fs, line 137: if the capture never started, calling `FS('shutdown')` won't fire this, right?

that is, in the following sequence of events, we won't call the start callback right?
1. `_fs_capture_on_startup` = false
2. call `FS('shutdown')` (assuming it won't fire shutdown, otherwise it would set `hadFsShutdown`)
3. call `FS('start')`, if it fired shutdown, then we could potentially be creating two observers?

src/observer.ts
```diff
@@ -661,4 +698,22 @@ export class DataLayerObserver {
...
+  /**
+   * Stops all handlers, removes monitors for their targets, and clears this observer.
+   * After this call, deferred rule registration (waitUntil timers, sleep retries, DOMContentLoaded)
+   * and {@link DataLayerObserver.registerRule} / {@link DataLayerObserver.registerOperator} are ignored;
+   * {@link DataLayerObserver.registerTarget} logs and returns undefined.
+   * Does not unregister custom operators on the class instance.
+   */
+  destroy(): void {
+    this.destroyed = true;
```
eugene-chang-fs, line 710: from claude:

```
destroy() does not reset listeners: DataLayerObserver.listeners is a map of event listeners added to window.
  destroy() stops handlers and removes monitors but doesn't clean up this.listeners. Whether this matters depends on
  whether listeners are the source of truth or just bookkeeping — worth checking.
```

i couldn't find how the public member `listeners` are used but it does look like claude's point is valid

src/embed/init.impl.ts
```diff
@@ -94,6 +99,80 @@ function _dlo_collectRules(): any[] {
   }
 }
 
+function readDloConfigFromWindow(win: { [key: string]: any }, rules: any[]): DataLayerConfig {
```
eugene-chang-fs, line 102: from claude:
```
_dlo_collectRules() on restart reads from window only: On restart, readDloConfigFromWindow picks up whatever
  _dlo_rules is set on window at restart time. If rules were dynamically added via registerRule (not via
  window._dlo_rules), they'd be lost. This is probably acceptable behavior for v1 of the feature but could be a footgun.
```
i also do think this is a valid point: on restart, any rules that were added via `registerRule` (i'm not sure how many in the wild there are, but still) would be lost

src/handler.ts
```diff
@@ -274,6 +274,10 @@ export default class DataHandler {
    * Stops listening for data layer changes or function calls.
    */
   stop() {
+    if (typeof this.timeoutId === 'number') {
+      window.clearTimeout(this.timeoutId);
+      this.timeoutId = null;
+    }
     window.removeEventListener(createEventType(this.source, this.target.path), this.listener as EventListener);
```
eugene-chang-fs, line 281: this isn't new to your change but i think we should guard against when `this.listener` is null/undefined, in case `stop` is called before `start`

src/embed/init.impl.ts
```diff
@@ -1,7 +1,7 @@
-/* eslint-disable no-underscore-dangle, camelcase */
+/* eslint-disable no-underscore-dangle, camelcase, no-param-reassign */
```
eugene-chang-fs, line 1: NIT: is this needed? i didn't see code that modified params

eugene-chang-fs (approved):
lgtm but couple of nits and couple of potential issues claude mentioned; i think warrants a review

src/embed/init.impl.ts
ScottLNorvell: Needed for reassignment of `win._dlo_observer`

image

van-fs:
Posting recap of understanding.  The sequence of events is something like the following:

1. FS.js capture starts.
2. DLO will readOnLoad
3. As eVar/prop variable assignments occur, DLO will fire and call FS APIs.
4. At some point user is inactive, fs.js stops capture.
5. eVar/prop assignments continue to occur.
6. User goes active and capture restarts (as a new session).
7. we want DLO to readOnLoad the data layer because eVars have changed

The nuance in step 7 is that a new session has started, and the end customer _wants_ the entire data layer contents.  A reason for this is that the entire data layer is needed to properly construct a segment/metric etc.  Because DLO doesn't have a concept of inactive timeout, it doesn't handle this scenario.  DLO only retrieves a snapshot of the data layer when it _first_ loads.  Because this isn't its initial load, no snapshot for the new session.

Given that's the need, I'm more inclined to expose the `readOnLoad` function externally to the external JS that is proposed.  That way external code from DLO can monitor for new session restarts and if needed tell DLO to re-read the data layer.  This would eliminate the teardown/recreation activity, state tracking for destroy, etc.  We should be able to simply call a very proven path (readOnLoad) versus creating new code paths in DLO internals.

PR #248, opened 2026-05-08
[VAL-10160] replay readOnload on fs restart

ScottLNorvell:
Another approach to DLO "restart" functionality: Only replay readOnLoad when the flag is set and FS restarts.

Relates to "full" restart approach here: https://github.com/fullstorydev/fullstory-data-layer-observer/pull/247


src/handler.ts
```diff
@@ -30,10 +30,11 @@ export default class DataHandler {
    * @param target in the data layer
    * @param debug true optionally enables debugging data transformation (defaults to console.debug)
    * @param debounce number of milliseconds to debounce property value assignments (defaults to 250ms)
+   * @param readOnLoad when true, initial read and FullStory lifecycle replay should emit the current snapshot for this handler
    * @throws will throw an error if the data layer is not found (i.e. undefined or null)
    */
   constructor(private readonly source: string, public readonly target: DataLayerValue, public debug = false,
-    public debounce = DataHandler.DefaultDebounceTime) {
+    public debounce = DataHandler.DefaultDebounceTime, public readonly readOnLoad = false) {
```
van-fs: readOnLoad is typically what we want to favor.  The `DataLayerObserver` constructor in init.impl.ts sets this to `true`.  Suggest changing the optional so it defaults to `true`.

src/observer.ts
```diff
@@ -370,36 +376,11 @@ export class DataLayerObserver {
...
-        } catch (err) {
-          Logger.getInstance().error(LogMessageType.ObserverReadError,
-            {
-              path: workingTarget.path,
-              reason: err.message,
-            });
-          Telemetry.error(errorType.observerReadError);
-        }
-      }
+    if (readOnLoad) {
```
van-fs: What's the idiomatic way to do this?  If the handler is self-aware of readOnLoad because it's passed in to the constructor, couldn't you just call `handler.readOnLoad()` and expect the appropriate behavior.  Would the observer need to know when to condition and call into the handler?

src/handler.ts
```diff
@@ -51,6 +52,42 @@ export default class DataHandler {
     }
   }
 
+  /**
+   * Re-emits the initial-load events for this handler, mirroring the read-on-load semantics
+   * used at registration time. When `initialValue` is an array, each element is fanned out
+   * through `fireEvent`. Otherwise, for object-typed targets, `fireEvent()` is invoked once,
+   * which queries the current snapshot. Other types (e.g. functions) are a no-op.
+   */
+  fireReadOnLoad(initialValue?: any): void {
```
van-fs: Nit: could we just call it `readOnLoad`.  That way it matches the the `readOnLoad` configuration properties.

src/observer.ts
ScottLNorvell: That's a good point. I was on the fence 🤺 but we could honor this flag inside of readOnLoad and move it out of the observer concern. Often in fs.js we would add `maybe` to the function name so `maybeReadOnLoad` just so it's clear we are honoring some internal logic to ignore.

I'll make the update to internalize the flag (moving it out of observer 👍 )

src/handler.ts
ScottLNorvell: Oh interesting. I suppose it doesn't need to be optional since we're always passing it in. I'll default it to `true`, see what it does to the tests, and go from there 👍

ScottLNorvell: Defaulting to true === all tests still pass 👍

src/observer.ts
ScottLNorvell: Updated!

src/handler.ts
ScottLNorvell: Would cause a naming conflict with the config value passed in. I updated to `maybeReadOnLoad` per above 😄

van-fs (approved):
LGTM

PR #250, opened 2026-05-28
refactor: route _fs_namespace reads through getFsNamespace helper

pablofullstory:
## Summary

DLO had four hard-coded `window._fs_namespace` reads, which silently break for
sites whose FullStory snippet uses a custom namespace (set via the
`data-fs-namespace` script attribute) without also setting the `_fs_namespace`
global. This PR centralizes the lookup behind a single helper that mirrors the
resolution order used by `@fullstory/browser-api`.

- Add `src/utils/fsNamespace.ts` exporting `getFsNamespace(win = window)`:
  `data-fs-namespace` on `document.currentScript` → `_fs_namespace` global → `'FS'`.
  Self-contained (no `@fullstory/browser-api` dependency); defensive `try/catch`
  around `document` access; empty-string attribute falls through to the global.
- Replace the four direct reads with helper calls:
  - `src/utils/logger.ts` — `FullStoryAppender.log`
  - `src/operators/fsApi/fsApi.ts` — `FSApiOperator.handleData`
    (also tightens the throw message to `Fullstory namespace is not a function`)
  - `src/target.ts` — `DataLayerTarget` constructor's "is this the FS object?" guard
  - `src/embed/init.impl.ts` — `attachDloFullStoryLifecycle`

### Behavior

- Existing call sites that only had the `_fs_namespace` global still work — the
  helper falls back to the global, which is the same floor the old code had.
- Sites that set `data-fs-namespace="MyFS"` on the FullStory snippet but do not
  set `_fs_namespace` now resolve correctly (this was the bug).
- `document.currentScript` is only meaningful while a script tag is being
  evaluated synchronously. DLO calls the helper from deferred callbacks
  (appender log, operator `handleData`, lifecycle hook), so the attribute path
  is opportunistic; the global remains the practical fallback. This matches how
  `@fullstory/browser-api`'s own `getFsNamespace` is shaped.

## Test plan

- [x] `npm test` — 428 passing, including 8 new cases in `test/fs-namespace.spec.ts`
      covering: attribute wins over global, empty-string attribute is ignored,
      missing/null `currentScript` falls through, errors during document access
      are swallowed, default `'FS'` floor, and `undefined` window.
- [x] `npm run lint` — clean (only pre-existing `no-console` warnings remain).
- [x] `npm run build` — clean.
- [x] Existing specs that seed `window._fs_namespace = 'FS'` (`logger.spec.ts`,
      `lifecycle-embed.spec.ts`, `operator-fsApi.spec.ts`) still pass via the
      helper's global fallback.

---

> [!NOTE]
> **Low Risk**
> Localized refactor with global fallback preserved; behavior change only when `_fs_namespace` was unset but a custom script namespace was configured.
> 
> **Overview**
> Introduces **`getFsNamespace`** so DLO resolves the FullStory client the same way as fs.js: `data-fs-namespace` on `document.currentScript` (when available), then `window._fs_namespace`, then **`'FS'`**. That fixes integrations that only set a custom namespace via the script attribute and never set the global.
> 
> **Four call sites** now use the helper instead of reading `_fs_namespace` directly: `FullStoryAppender`, `FSApiOperator.handleData`, the `DataLayerTarget` FS-object guard, and `attachDloFullStoryLifecycle`. The FS API operator error message is updated to *Fullstory namespace is not a function*.
> 
> Adds **`test/fs-namespace.spec.ts`** with eight cases for precedence, fallbacks, empty attribute, document errors, and missing `window`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24894f2ad9f48232eac719c56fcc1d3e5501cff0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/utils/fsNamespace.ts
+++ b/src/utils/fsNamespace.ts
@@ -1,0 +1,28 @@
+/* eslint-disable no-underscore-dangle, import/prefer-default-export */
+
+/**
+ * Resolves the FullStory client namespace on `window`.
+ *
+ * Mirrors the resolution order used by `@fullstory/browser-api`:
+ *   1. The `data-fs-namespace` attribute on `document.currentScript` (best-effort).
+ *   2. The `_fs_namespace` global set on `window` by the FullStory snippet.
+ *   3. The default namespace `'FS'`.
+ *
+ * Note: `document.currentScript` is only meaningful while the helper module is being
+ * evaluated synchronously by a script tag. DLO calls this helper from deferred callbacks
+ * (appender log, operator handleData, lifecycle hook), so the attribute lookup is
+ * opportunistic and the global is the practical fallback floor — same behavior as the
+ * existing `(window as any)._fs_namespace` reads it replaces.
+ *
+ * @param win the global to read from; defaults to `window`
+ */
+export function getFsNamespace(win: any = window): string {
+  try {
+    const script = (win && win.document && win.document.currentScript) as HTMLScriptElement | null;
+    const attr = script && script.getAttribute && script.getAttribute('data-fs-namespace');
+    if (attr) return attr;
+  } catch (_) {
+    // ignore: cross-origin or detached document access can throw
+  }
+  return (win && win._fs_namespace) || 'FS';
+}
--- a/src/utils/fsNamespace.ts
+++ b/src/utils/fsNamespace.ts
@@ -3,7 +3,7 @@
 /**
  * Resolves the FullStory client namespace on `window`.
  *
- * Mirrors the resolution order used by `@fullstory/browser-api`:
+ * Mirrors the resolution order used by fs.js:
  *   1. The `data-fs-namespace` attribute on `document.currentScript` (best-effort).
  *   2. The `_fs_namespace` global set on `window` by the FullStory snippet.
  *   3. The default namespace `'FS'`.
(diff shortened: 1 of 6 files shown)


src/utils/fsNamespace.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable no-underscore-dangle, import/prefer-default-export */
+
+/**
+ * Resolves the FullStory client namespace on `window`.
+ *
+ * Mirrors the resolution order used by `@fullstory/browser-api`:
```
ScottLNorvell: let's say `fs.js` instead of `@fullstory/browser-api` since that's an _internal_ package and this is theoretically an open source repo.

test/fs-namespace.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+describe('getFsNamespace', () => {
+  it('returns the data-fs-namespace attribute when set on document.currentScript', () => {
+    const win = makeWin({ scriptAttr: 'CustomFS', fsNamespace: 'FS' });
+    expect(getFsNamespace(win)).to.eq('CustomFS');
+  });
+
+  it('prefers the script attribute over _fs_namespace when both are present', () => {
+    const win = makeWin({ scriptAttr: 'AttrWins', fsNamespace: 'GlobalLoses' });
```
ScottLNorvell, line 51: duck_hunt_laugh

ScottLNorvell (approved):
LGTM! Just the one comment about removing the reference to `@fullstory/browser-api`. (which I'm sure that silly LLM put there hahhaha)

PR #252, opened 2026-06-17
feat: cache FullStory namespace from the executing script tag at load

pablofullstory:
## Summary
- `document.currentScript` is only non-null while a script runs synchronously and is reset to `null` once control returns to the event loop, so DLO's deferred callbacks (appender log, operator `handleData`, lifecycle hook) never see it. The attribute path never hit there, so DLO fell back to the `_fs_namespace` global.
- FullStory's loader now stamps the DLO library `<script>` tag with `data-fs-namespace`. At module-evaluation time (when this bundle first runs synchronously) `document.currentScript` is that stamped script, so we read the attribute once then, while it is live, and cache the resolved **string** (not the element, so it can still be garbage collected after DOM removal).
- Deferred callbacks now resolve a custom namespace without relying on the `_fs_namespace` global. Backward-safe: when the script tag isn't stamped the cache is empty and resolution falls back to the global, same as today.

## Resolution order (`getFsNamespace`)
1. `data-fs-namespace` on the live `document.currentScript`.
2. The same attribute captured at module-evaluation time (`cachedScriptNamespace`).
3. The `_fs_namespace` global.
4. Default `'FS'`.

## Changes
- `src/utils/fsNamespace.ts`: module-load cache of the namespace from `document.currentScript`.
- `test/fs-namespace.spec.ts`: added cases for the cached namespace (resolves when live `currentScript` is null; prefers cache over the global).
- Version bump `4.1.6` -> `4.1.7` + CHANGELOG entry.

## Test plan
- [x] `npx eslint` on changed files (clean)
- [x] `mocha` `test/fs-namespace.spec.ts` (10/10 passing)
- [ ] CI

## Follow-up
After merge, manually cut the `v4.1.7` release tag so it can be synced into the FullStory monorepo via `tools/opensource.go sync fullstory-data-layer-observer v4.1.7`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/utils/fsNamespace.ts
+++ b/src/utils/fsNamespace.ts
@@ -3,10 +3,16 @@
 const FS_NAMESPACE_ATTR = 'data-fs-namespace';
 const DEFAULT_NAMESPACE = 'FS';
 
-function readCurrentScriptNamespace(): string | undefined {
+/**
+ * Reads the data-fs-namespace attribute from `win.document.currentScript`.
+ * Returns undefined when there is no document, no current script, no attribute,
+ * or when access throws (e.g. cross-origin or detached document).
+ *
+ * @param win the global to read from
+ */
+function readCurrentScriptNamespace(win: any): string | undefined {
   try {
-    if (typeof document === 'undefined') return undefined;
-    const script = document.currentScript as HTMLScriptElement | null;
+    const script = (win && win.document && win.document.currentScript) as HTMLScriptElement | null;
     return (script && script.getAttribute && script.getAttribute(FS_NAMESPACE_ATTR)) || undefined;
   } catch (_) {
     return undefined;
@@ -22,7 +28,7 @@ function readCurrentScriptNamespace(): string | undefined {
  * currentScript is still live and cache the resolved string (not the element, so it
  * can still be garbage collected after DOM removal).
  */
-const cachedScriptNamespace = readCurrentScriptNamespace();
+const cachedScriptNamespace = readCurrentScriptNamespace(typeof window !== 'undefined' ? window : undefined);
 
 /**
  * Resolves the FullStory client namespace on `win`.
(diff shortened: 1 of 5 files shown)


src/utils/fsNamespace.ts
```diff
@@ -1,28 +1,49 @@
...
+    fromLiveScript = (script && script.getAttribute && script.getAttribute(FS_NAMESPACE_ATTR))
+      || undefined;
   } catch (_) {
-    // ignore: cross-origin or detached document access can throw
+    // ignore: cross-origin or detached document access can throw. cachedScriptNamespace
+    // is captured at module load and does not depend on win.document, so it is still
+    // consulted below before the global fallback.
   }
-  return (win && win._fs_namespace) || 'FS';
+  return fromLiveScript || cachedScriptNamespace || (win && win._fs_namespace) || DEFAULT_NAMESPACE;
```
ScottLNorvell: The one thing I'd say here is I'm not sure things are going to change much so it's probably ok to skip checking the script _again_ and just return: 

```js
return cachedScriptNamespace || (win && win._fs_namespace) || DEFAULT_NAMESPACE
```

Or (since it's not really hurting anyone) you can simplify to:

```js
return readCurrentScriptNamespace(win) || cachedScriptNamespace || (win && win._fs_namespace) || DEFAULT_NAMESPACE
```

(that would require having a `win` arg that defaults to window 🤷 )

ScottLNorvell (approved):
Cleanup suggestion otherwise LGTM 👍 

web-capture-stamp

src/utils/fsNamespace.ts
pablofullstory: done in [6768893](https://github.com/fullstorydev/fullstory-data-layer-observer/pull/252/commits/676889323e0a727eff1b2993e0607a1f6ff3141a)

